PR #16, opened 2024-07-18
Feat: use consistent_df package 

AlexTheWizardL:
This PR covers changes to integrate the `consistent_df` package and use the `enforce_dtypes()` method to ensure robust type consistency. All methods that previously used the `astype()` approach have been replaced with `enforce_dtypes()`. Additionally, a constants file has been created to contain all field types.

@lasuk Please review the changes and provide feedback or approval for merging.

### DIFF
--- a/pyledger/constants.py
+++ b/pyledger/constants.py
@@ -1,0 +1,47 @@
+"""Constants used throughout the application."""
+
+FX_ADJUSTMENTS_COLUMNS = {
+    "date": "datetime64[ns, Europe/Berlin]",
+    "adjust": "string[python]",
+    "credit": "Int64",
+    "debit": "Int64",
+    "text": "string[python]",
+}
+
+VAT_CODE_COLUMNS = {
+    "id": "string[python]",
+    "date": "datetime64[ns, Europe/Berlin]",
+    "text": "string[python]",
+    "inclusive": "bool",
+    "account": "Int64",
+    "inverse_account": "Int64",
+    "rate": "Float64",
+}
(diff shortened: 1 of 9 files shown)


pyledger/constants.py
```diff
@@ -0,0 +1,47 @@
...
+    "date": "datetime64[ns, Europe/Berlin]",
+    "account": "Int64",
+    "counter_account": "Int64",
+    "currency": "string[python]",
+    "amount": "Float64",
+    "base_currency_amount": "Float64",
+    "vat_code": "string[python]",
+    "text": "string[python]",
+    "document": "string[python]",
+}
```
lasuk, line 40: There are two further constants related to ledger columns that should IMO be defined in the same place:

```
    LEDGER_COLUMN_SHORTCUTS = {
        "cur": "currency",
        "vat": "vat_code",
        "target": "target_balance",
        "base_amount": "base_currency_amount",
        "counter": "counter_account",
    }
    LEDGER_COLUMN_SEQUENCE = [
        "id", "date", "account", "counter_account", "currency", "amount",
        "target_balance", "balance", "base_currency_amount",
        "base_currency_balance", "vat_code", "text", "document",
    ]
```
(Applies to all data frame types)

lasuk (commented):
Thanks for switching our code to using the new `enforce_dtypes` method.

There is some more work to do to refactor the pre-existing constants into a coherent set of new constants, rather than having old and new constants live side by side. 

I left a few comments.

Issue #2, opened 2024-05-15
Drop DataFrame Index in `LedgerEngine.standardize_...()` methods


lasuk:
`LedgerEngine.standardize_vat_codes()` currently returns a DataFrame with the vat_code as index. 

We decided to stop using DataFrame indexes and should adapt the interface accordingly.

This change will affect the test suite in cashctrl_ledger. See [example](https://github.com/macxred/cashctrl_ledger/blob/6b8587c8641d8d893ec886f969eb755685413bac/tests/test_vat_code.py#L26-L28).

PR #18, opened 2024-08-14
Feat: implement round_series_to_precision() method and tests for it

AlexTheWizardL:
### This PR introduces the `round_series_to_precision()` method, which is necessary for implementing [issue #35](https://github.com/macxred/cashctrl_ledger/issues/35).

Key points:

- Python does not natively support method overloading, so instead of overloading the existing `round_to_precision()` method, I created a new method, `round_series_to_precision()`, for the vectorized version. This avoids adding type checks and splitting the logic within a single method.
- The tests are currently written under the `TestLedger()` class. These tests are expected to be refactored in [milestone #2](https://github.com/macxred/pyledger/milestone/2).

@lasuk Please review the changes and provide feedback or approval for merging

### DIFF
--- a/pyledger/ledger_engine.py
+++ b/pyledger/ledger_engine.py
@@ -924,45 +924,57 @@ class LedgerEngine(ABC):
         """
 
     def round_to_precision(
-        self, amount: float, ticker: str, date: datetime.date = None
-    ) -> float:
-        """Round an amount to the precision of a specified asset or currency.
-
-        This method retrieves the precision for the specified ticker and date,
-        then rounds the amount to the nearest multiple of this precision.
-
-        Args:
-            amount (float): Value to be rounded.
-            ticker (str): Ticker symbol of the asset or currency.
-            date (datetime.date, optional): Date for precision determination.
-                                            Defaults to today's date.
-
-        Returns:
-            float: Rounded amount, adjusted to the specified asset's precision.
-        """
-        precision = self.precision(ticker=ticker, date=date)
-        result = round(amount / precision, 0) * precision
-        return round(result, -1 * math.floor(math.log10(precision)))
-
-    def round_series_to_precision(
         self,
-        amount: pd.Series,
-        currency: pd.Series,
-    ) -> pd.Series:
-        """Round the amounts in the series to the specified precision based on the currency.
+        amount: float | list | pd.Series,
+        ticker: str | list | pd.Series,
+        date: datetime.date = None
+    ) -> float | list:
+        """Round amounts to the precision of the specified ticker (currency or asset) using
+        Python lists or Pandas Series.
 
         Args:
-            amount (pd.Series): Series of amounts to round.
-            currency (pd.Series): Series of currency codes corresponding to each amount.
+            amount (float, list, or pd.Series): Value(s) to be rounded. If a list or pd.Series,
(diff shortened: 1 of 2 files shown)


pyledger/ledger_engine.py
```diff
@@ -944,6 +944,26 @@ def round_to_precision(
...
+    def round_series_to_precision(
+        self,
+        amount: pd.Series,
+        currency: pd.Series,
+    ) -> pd.Series:
+        """Round the amounts in the series to the specified precision based on the currency.
+
+        Args:
+            amount (pd.Series): Series of amounts to round.
+            currency (pd.Series): Series of currency codes corresponding to each amount.
```
lasuk, line 956: Please explicitly state that the two series need to be of same length.

tests/test_standalone_ledger.py
```diff
@@ -138,3 +138,65 @@ def test_validate_account_balance(account: int, expected_length: int, expected_b
...
+            pd.Series([100.234, 200.567, 300.891, 400.789]),
+            pd.Series(["USD", "CHF", "CAD", "HKD"]),
+            pd.Series([100.23, 200.57, 300.89, 400.79])
+        ),
+    ]
+)
+def test_rounding(amounts, currencies, expected):
+    ledger = TestLedger()
+    result = ledger.round_series_to_precision(amounts, currencies)
+    pd.testing.assert_series_equal(result, expected)
```
lasuk, line 170: I do not see a benefit of using the additional complication of a parametrised test over a simpler test with hard-coded values.

(applies also to the next test below)

tests/test_standalone_ledger.py
lasuk, line 1: Please also test the following edge cases:
- Zero-length arguments
- Arguments of differing length raises an error (with sufficiently clear error message)

lasuk, line 1: Please start the test with a fixture that explicitly establishes a context for the expected behaviour:
Set precision to the expected values for all currencies tested and do not rely on an implicit default value.

lasuk (commented):
Thanks for implementing a vectored `precision` method. I left a few review comments.

As a general rule for working with vectored arguments: 
- Always test for zero-length input as an edge case.
- Keep in mind that the vector length might also be an input that needs testing. In our case, both vectors need to be of same length.

lasuk:
I have changed my opinion on the best way to combine the two related methods dealing with singleton and vector arguments.

Currently, we have one method taking singleton and another method taking pd.Series as arguments:
```
round_to_precision(self, amount: float, currency: str) -> float
round_series_to_precision(self, amount: pd.Series, currency: pd.Series) -> pd.Series
```

I propose instead to implement a single method where each argument can be either a singleton or a vector. The return value is a float if both arguments are singletons and a vector of float otherwise.
```
round_to_precision(self, amount: float | "vector of float", currency: str | "vector of str") -> float | vector of float
```

Motivation: The proposed solution is more flexible, it allows to provide one singleton and one vector as argument.

lasuk:
Also, the method shall be open to any vector input in python, including arrays, rather than narrowly focusing on a pd.Series. This must be a quite common problem in Pyhton and we shall stick with the usual pattern used in python. We should also stick to the usual return type for such cases (array instead of pd.Series?).

PR #19, opened 2024-08-16
Feat: create memory ledger

AlexTheWizardL:
### This PR covers changes of creating the MemoryLedger class, which extends StandaloneLedger to create a non-persistent ledger system.

This PR is better reviewed commit by commit:

- 71b2dfed81ca7cddcf9dc05409f5b62cb6764bdd - add initial structure of new class
- 3544e729dcc7de68264237665a47eaaae0585a35 - add accessor mutator functions for vat codes and corresponding tests
- 46b0ebf6dc8758a3d7386de067c76ac6adf78b85 - add accessor mutator functions for account charts and corresponding tests
- c6ada998353adba543fb6b5e68e34933df7573f6 - add accessor mutator functions for ledger and corresponding tests
- 2a83e16cf065f54c34bd3531df80a969adce1a2d - covers changes described in #2 
- 1ce9608bbd01f91e1cc20de1c8f1d97b56c47ee7 - add mirroring logic for VAT, accounts, and ledger with corresponding tests

@lasuk Please review the changes and provide feedback or approval for merging


pyledger/memory_ledger.py
```diff
@@ -0,0 +1,325 @@
...
+                                    present in the target DataFrame.
+        """
+        target_df = self.standardize_ledger(target)
+
+        if delete:
+            self._ledger = target_df
+        else:
+            self._ledger = self.standardize_ledger(
+                pd.concat([self._ledger, target_df]).drop_duplicates()
+            )
```
lasuk, line 323: The current logic only drops 'identical' duplicates. Our implementation in cashctrl_ledger is smarter: it drops duplicates that are financially equivalent.

There is no need for action at this time. The two implementations will automatically align as part of #25 which is scheduled for completion in October.

pyledger/memory_ledger.py
```diff
@@ -0,0 +1,323 @@
...
+            code (str): The VAT code to be added.
+            rate (float): The VAT rate, must be between 0 and 1.
+            account (str): The account identifier to which the VAT is applied.
+            inclusive (bool, optional): Determines whether the VAT is calculated as 'NET'
+                                        (True, default) or 'GROSS' (False). Defaults to True.
+            text (str, optional): Additional text or description associated with the VAT code.
+
+        Raises:
+            ValueError: If the VAT code already exists in the memory.
+        """
```
lasuk, line 71: IMO, this docstring only slightly differs from the abstract base class' docstring.
In this case, we should remove it entirely to inherit the docstring from the abstract bas class.

(Applies to all methods in the MemoryLedger class.)

lasuk (commented):
Thanks a lot for implementing the MemoryLedger class and for providing initial implementations on the abstract base classes.

I have left a series of comments and have pushed two commits with changes to documents and defaults to this PR. My comments focus on the new MemoryLedger class, and I deliberately did not comment on the abstract base tests, as those will be the subject of Tuesday's brainstorming session.

Back to you @AlexTheWizardL

pyledger/memory_ledger.py
```diff
@@ -0,0 +1,325 @@
...
+            "GBP": 0.01,
+            "JPY": 1.00,
+            "NZD": 0.01,
+            "NOK": 0.01,
+            "SEK": 0.01,
+            "USD": 0.01,
+        },
+    }
+
+    def __init__(self) -> None:
```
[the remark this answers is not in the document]
AlexTheWizardL: Make sense to me

pyledger/memory_ledger.py
```diff
@@ -0,0 +1,325 @@
...
+            rate (float): The VAT rate, must be between 0 and 1.
+            account (str): The account identifier to which the VAT is applied.
+            inclusive (bool, optional): Determines whether the VAT is calculated as 'NET'
+                                        (True, default) or 'GROSS' (False). Defaults to True.
+            text (str, optional): Additional text or description associated with the VAT code.
+
+        Raises:
+            ValueError: If the VAT code is not found in the memory.
+        """
+        if not self._vat_codes["id"].isin([code]).any():
```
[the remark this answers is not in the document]
AlexTheWizardL: Accessor, mutator, and mirror methods ensure there is no duplicate ID. Duplicate IDs can be possible only by accessing variables directly.

We can ensure there is only one matching row with:
```python
        if self._vat_codes["id"].isin([code]).sum() == 1:
            raise ValueError(f"VAT code '{code}' already exists")
```

But in this scenario, if somehow there are already 2 or more VAT codes we'll add one more to the existing entries instead of throwing an error as currently did.

Your thoughts?

pyledger/memory_ledger.py
```diff
@@ -0,0 +1,325 @@
...
+            delete (bool, optional): If True, deletes VAT codes on the remote account
+                                     that are not present in target_state.
+        """
+        target_df = self.standardize_vat_codes(target)
+
+        if delete:
+            self._vat_codes = target_df
+        else:
+            self._vat_codes = self.standardize_vat_codes(
+                pd.concat([self._vat_codes, target_df]).drop_duplicates()
```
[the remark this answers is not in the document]
AlexTheWizardL: I conducted an investigation into the behavior of `drop_duplicates` using two DataFrames, `df1` and `df2`, which contain some overlapping rows.

In the case of duplicating rows, the data in those rows is identical, which is why the only observable difference between `keep="first"` and `keep="last"` is the order of duplicated rows and their default DataFrame index. The content of the rows remains unchanged, so this behavior is useful when you need to preserve either the earlier or later occurrence of rows while cleaning up duplicates, without affecting the underlying data. 

<details>
<summary>Investigation: `drop_duplicates` Behavior on Concatenated DataFrames</summary>

#### DataFrames:
**df1:**
| id   | text  | inclusive | account | rate |
|------|-------|-----------|---------|------|
| A001 | Item 1 | True      | 1001    | 0.05 |
| A002 | Item 2 | True      | 1001    | 0.05 |
| A003 | Item 3 | True      | 1001    | 0.05 |
| A004 | Item 4 | True      | 1001    | 0.05 |

**df2:**
| id   | text  | inclusive | account | rate |
|------|-------|-----------|---------|------|
| A005 | Item 5 | True      | 1001    | 0.05 |
| A002 | Item 2 | True      | 1001    | 0.05 |
| A006 | Item 6 | True      | 1001    | 0.05 |
| A007 | Item 7 | True      | 1001    | 0.05 |

### Steps:
1. **Concatenated `df1` and `df2`:** I combined the two DataFrames using `pd.concat()`, resulting in a larger DataFrame with some duplicated rows based on the `"id"` and `"text"` fields.
2. **Applied `drop_duplicates`:**
   - **`keep="first"`**: Retained the first occurrence of each duplicated row.
   - **`keep="last"`**: Retained the last occurrence of each duplicated row.

### Results:
- **`keep="first"`**:
    The first occurrence of each duplicated row is kept, and duplicates are dropped. The order remains as it appears in the combined DataFrame.

#### Keep First:
| id   | text  | inclusive | account | rate |
|------|-------|-----------|---------|------|
| A001 | Item 1 | True      | 1001    | 0.05 |
| A002 | Item 2 | True      | 1001    | 0.05 |
| A003 | Item 3 | True      | 1001    | 0.05 |
| A004 | Item 4 | True      | 1001    | 0.05 |
| A005 | Item 5 | True      | 1001    | 0.05 |
| A006 | Item 6 | True      | 1001    | 0.05 |
| A007 | Item 7 | True      | 1001    | 0.05 |
  
- **`keep="last"`**:
    The last occurrence of each duplicated row is retained, and earlier duplicates are dropped. This changes the default DataFrame index but the content is the same.

#### Keep Last:
| id   | text  | inclusive | account | rate |
|------|-------|-----------|---------|------|
| A001 | Item 1 | True      | 1001    | 0.05 |
| A003 | Item 3 | True      | 1001    | 0.05 |
| A004 | Item 4 | True      | 1001    | 0.05 |
| A005 | Item 5 | True      | 1001    | 0.05 |
| A002 | Item 2 | True      | 1001    | 0.05 |
| A006 | Item 6 | True      | 1001    | 0.05 |
| A007 | Item 7 | True      | 1001    | 0.05 |


### Code Snippet and Resulting DataFrames:

```python
import pandas as pd

# Define DataFrames
df1 = pd.DataFrame({
    "id": pd.Series(["A001", "A002", "A003", "A004"], dtype="string[python]"),
    "text": pd.Series(["Item 1", "Item 2", "Item 3", "Item 4"], dtype="string[python]"),
    "inclusive": pd.Series([True, True, True, True], dtype="bool"),
    "account": pd.Series([1001, 1001, 1001, 1001], dtype="Int64"),
    "rate": pd.Series([0.05, 0.05, 0.05, 0.05], dtype="Float64")
})

df2 = pd.DataFrame({
    "id": pd.Series(["A005", "A002", "A006", "A007"], dtype="string[python]"),
    "text": pd.Series(["Item 5", "Item 2", "Item 6", "Item 7"], dtype="string[python]"),
    "inclusive": pd.Series([True, True, True, True], dtype="bool"),
    "account": pd.Series([1001, 1001, 1001, 1001], dtype="Int64"),
    "rate": pd.Series([0.05, 0.05, 0.05, 0.05], dtype="Float64")
})

# Concatenate df1 and df2
combined_df = pd.concat([df1, df2])

# Drop duplicates with 'keep="first"' and 'keep="last"'
df_keep_first = combined_df.drop_duplicates(keep="first")
df_keep_last = combined_df.drop_duplicates(keep="last")

# Display resulting DataFrames
print("Keep First:")
print(df_keep_first)

print("\nKeep Last:")
print(df_keep_last)
```

</details>

**Is keeping the last occurrences of the duplicate rows the behaviour you would expect?**

pyledger/memory_ledger.py
```diff
@@ -0,0 +1,325 @@
...
+        Args:
+            target (pd.DataFrame): DataFrame containing VAT rates in
+                                         the pyledger.vat_codes format.
+            delete (bool, optional): If True, deletes VAT codes on the remote account
+                                     that are not present in target_state.
+        """
+        target_df = self.standardize_vat_codes(target)
+
+        if delete:
+            self._vat_codes = target_df
```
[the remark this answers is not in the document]
AlexTheWizardL: Ensuring there are no duplicate rows in `target_df` is a smart move.

pyledger/memory_ledger.py
AlexTheWizardL: I asked ChatGPT and here is what he responds:

<details>
<summary>ChatGPT response</summary>

In the Python community, the general consensus around writing docstrings for class methods, particularly in subclass implementations, is governed by principles of **clarity** and **maintainability**. Here's what the community standards say, along with recommendations for your specific case involving `MemoryLedger` and `LedgerEngine`:

1. **Docstrings in Abstract Classes:**
   - **Abstract methods** in an abstract class (like `LedgerEngine`) should include comprehensive docstrings, as they provide the expected behavior and describe the contract for all subclasses.
   - This is important because the abstract class is often seen as the blueprint or interface for its subclasses, so providing clear documentation ensures the intended use is understood.

2. **Docstrings in Subclasses (like `MemoryLedger`):**
   - **No changes from parent class**: If a method in `MemoryLedger` is an implementation of a method from `LedgerEngine` and its behavior does not change in any meaningful way, Python community standards (such as **PEP 257**) generally suggest **not duplicating the docstring**. The reasoning is that redundant documentation can lead to inconsistencies and make maintenance harder.
     - However, you can inherit the docstring from the parent class, which is allowed and encouraged to avoid duplication. Python provides a way to do this using `__doc__` inheritance automatically.
     - If using an IDE or static analysis tool, they will often still provide the parent class's docstring.

     Example (no new docstring):
     ```python
     class MemoryLedger(LedgerEngine):
         def some_method(self):
             super().some_method()
             # Method logic follows the abstract class contract.
     ```

3. **Slight changes from parent class**: 
   - If the method's implementation in `MemoryLedger` **slightly modifies** the behavior or adds specific context to the base class, it’s recommended to include a **concise docstring** that highlights what is different, without duplicating the entire docstring.
   - You can reference the base class's docstring, noting only the modification.

     Example (minor modifications):
     ```python
     class MemoryLedger(LedgerEngine):
         def some_method(self):
             """Performs the same operation as LedgerEngine but with a custom cache handling."""
             super().some_method()
             # Additional logic here.
     ```

### Key Principles
- **Avoid redundancy**: Don’t repeat docstrings when behavior hasn’t changed.
- **Document modifications**: If there are changes, focus on documenting those, while potentially referring to the parent docstring for general behavior.
- **Comprehensive for abstract classes**: Ensure the abstract class has full documentation since it sets the expectations for subclasses.
</details>

In summary, if `MemoryLedger` does not introduce significant changes to the method logic from `LedgerEngine`, you should avoid adding redundant docstrings. Instead, inherit or rely on the parent class's documentation. If there are slight changes, document only what differs while referencing the base behavior.

tests/base_test_dump_restore.py
```diff
@@ -0,0 +1,93 @@
...
+    1,  2024-05-24,  1000,            1999,      CHF,     100.00, single entry,     file_1.txt
+    42, 2024-06-21,  1020,                ,      USD,     100.00, collective entry, file_2.txt
+    42, 2024-06-21,      ,            4000,      USD,      50.00, collective entry, file_2.txt
+    42, 2024-06-21,      ,            5000,      USD,      50.00, collective entry, file_2.txt
+"""
+
+
+LEDGER_ENTRIES = pd.read_csv(StringIO(LEDGER_CSV), skipinitialspace=True)
+ACCOUNTS = pd.read_csv(StringIO(ACCOUNT_CSV), skipinitialspace=True)
+VAT_CODES = pd.read_csv(StringIO(VAT_CSV), skipinitialspace=True)
```
lasuk, line 31: What do you think of moving hard coded test data into a separate unit to avoid redundancy?

tests/base_test_dump_restore.py
```diff
@@ -0,0 +1,93 @@
...
+        initial_ledger = ledger.ledger()
+        initial_vat_codes = ledger.vat_codes()
+        initial_account_chart = ledger.account_chart()
+
+        yield
+
+        # Restore initial state
+        ledger.mirror_ledger(initial_ledger, delete=True)
+        ledger.mirror_vat_codes(initial_vat_codes, delete=True)
+        ledger.mirror_account_chart(initial_account_chart, delete=True)
```
lasuk, line 53: Why do we need to restore the initial state?

tests/base_test_dump_restore.py
```diff
@@ -0,0 +1,93 @@
...
+            ledger.standardize_vat_codes(VAT_CODES), ledger.vat_codes(), ignore_index=True
+        )
+        assert_frame_equal(
+            ledger.standardize_account_chart(ACCOUNTS), ledger.account_chart(), ignore_index=True
+        )
+        expected_ledger = sorted(
+            ledger.txn_to_str(ledger.standardize_ledger(LEDGER_ENTRIES)).values()
+        )
+        ledger = sorted(ledger.txn_to_str(ledger.ledger()).values())
+        assert expected_ledger == ledger
```
lasuk, line 71: Better readable version of the same:
```
        target = ledger.txn_to_str(lLEDGER_ENTRIES).values()
        actual = ledger.txn_to_str(ledger.ledger()).values()
        assert sorted(target) == sorted(actual), "Targeted and actual ledger differ"
```
Instead of:
```
        expected_ledger = sorted(
            ledger.txn_to_str(ledger.standardize_ledger(LEDGER_ENTRIES)).values()
        )
        ledger = sorted(ledger.txn_to_str(ledger.ledger()).values())
        assert expected_ledger == ledger
```

tests/base_test_ledger.py
```diff
@@ -0,0 +1,286 @@
...
+    group, account, currency, vat_code, text
+    /Assets, 10021,      EUR,         , Test EUR Bank Account
+    /Assets, 10022,      USD,         , Test USD Bank Account
+    /Assets, 10023,      CHF,         , Test CHF Bank Account
+    /Assets, 19991,      EUR,         , Transitory Account EUR
+    /Assets, 19992,      USD,         , Transitory Account USD
+    /Assets, 19993,      CHF,         , Transitory Account CHF
+    /Assets, 19999,      CHF,         , Transitory Account CHF
+    /Assets, 22000,      CHF,         , Input Tax
+"""
```
lasuk, line 19: Let's use four digit account numbers (this is more common).

tests/test_dump_restore.py
```diff
@@ -0,0 +1,15 @@
+"""This module contains the test suite for dump and restore operations using the MemoryLedger
+implementation. It inherits the common account chart tests from the BaseTestDumpAndRestore
+abstract base class.
+"""
```
lasuk, line 4: Short and concise doc string, please.

tests/test_clear.py
```diff
@@ -0,0 +1,14 @@
+"""This module contains the test suite for clearing ledger system using the MemoryLedger
+implementation. It inherits the common clear test from the BaseTestClear abstract base class.
+"""
```
lasuk, line 3: Short and concise doc string, please.

tests/base_account_chart.py
```diff
@@ -0,0 +1,171 @@
...
+            "text": "test update account without VAT",
+            "vat_code": None,
+            "group": "/Assets/Anlagevermögen",
+        }
+        ledger.modify_account(**new_account)
+        updated_accounts = ledger.account_chart()
+        outer_join = pd.merge(account_chart, updated_accounts, how="outer", indicator=True)
+        created_accounts = outer_join[outer_join["_merge"] == "right_only"].drop("_merge", axis=1)
+
+        assert len(created_accounts) == 1, "Expected exactly one row to be added"
```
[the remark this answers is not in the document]
lasuk: (also applies to test to modify_vat_code; in modify_ledger, please verify the number of unique id's remains unchanged).

tests/base_vat_codes.py
```diff
@@ -0,0 +1,129 @@
...
+        ledger.mirror_vat_codes(VAT_CODES, delete=False)
+        mirrored_df = ledger.vat_codes()
+        standardized_df = ledger.standardize_vat_codes(VAT_CODES)
+        m = standardized_df.merge(mirrored_df, how="left", indicator=True)
+        assert (m["_merge"] == "both").all(), (
+            "Mirroring error: Some target VAT codes were not mirrored"
+        )
+
+        # Mirror target vat codes onto server with delete=True
+        ledger.mirror_vat_codes(VAT_CODES, delete=True)
```
lasuk, line 93: Please drop some rows from 'VAT_CODES' to test that deletion works even if the original ledger system was empty.

tests/base_vat_codes.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def test_mirror(self, ledger):
+        # Mirror test vat codes onto server with delete=False
+        ledger.mirror_vat_codes(VAT_CODES, delete=False)
+        mirrored_df = ledger.vat_codes()
+        standardized_df = ledger.standardize_vat_codes(VAT_CODES)
+        m = standardized_df.merge(mirrored_df, how="left", indicator=True)
+        assert (m["_merge"] == "both").all(), (
+            "Mirroring error: Some target VAT codes were not mirrored"
+        )
```
lasuk, line 90: Is there not an easier or more explicit way to test that data frames are identical?

(applies throughout the module)

tests/base_account_chart.py
```diff
@@ -0,0 +1,171 @@
...
+            ledger.add_account(
+                account=1200,
+                currency="EUR",
+                text="test account",
+                vat_code="TestCodeAccounts",
+                group="/Assets/Anlagevermögen",
+            )
+        add_account()
+        with pytest.raises(ValueError):
+            add_account()
```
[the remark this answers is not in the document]
lasuk: (applies also to vat_codes)

tests/base_account_chart.py
```diff
@@ -0,0 +1,171 @@
...
+        def add_account():
+            ledger.add_account(
+                account=1200,
+                currency="EUR",
+                text="test account",
+                vat_code="TestCodeAccounts",
+                group="/Assets/Anlagevermögen",
+            )
+        add_account()
+        with pytest.raises(ValueError):
```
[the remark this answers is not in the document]
lasuk: (applies to all tests where we check that an error was raised)

tests/base_test_ledger.py
```diff
@@ -0,0 +1,286 @@
...
+
+class BaseTestLedger(ABC):
+
+    @abstractmethod
+    @pytest.fixture
+    def ledger(self):
+        pass
+
+    @pytest.mark.parametrize(
+        "ledger_id", set(LEDGER_ENTRIES["id"].unique()).difference([15, 16, 17, 18])
```
lasuk, line 85: Let's loop over all id's.
Excluding ids 15..18 is only necessary for CashCtrlLedger.

pyledger/__init__.py
```diff
@@ -22,7 +22,15 @@
 from .ledger_engine import LedgerEngine
 from .standalone_ledger import StandaloneLedger
 from .test_ledger import TestLedger
+from .memory_ledger import MemoryLedger
 from .text_ledger import TextLedger
 from .helpers import *
 from .time import *
 from .import constants
+from tests import (
```
lasuk, line 30: Is my understanding correct: We need to import BaseTest... here to make the tests accessible in cashctrl_ledger (or other derived packages)?

tests/__init__.py
lasuk, line 1: What is the benefit of having this file?

lasuk (commented):
Thanks for implementing the MemoryLedger class and BaseTests. 
It's a pleasure to see our how concept shapes up very nicely when implemented!

I added a commit to beautify doc strings, comments etc. and left a series of review comments. Back to you, @AlexTheWizardL

pyledger/memory_ledger.py
lasuk: Since the current logic is not correct, we could raise an error until #25 is implemented:
```python
            raise NotImplementedError(
                "mirror_ledger(..., delete=False) will be implemented as part of #25.")
```
Instead of:
```python
            self._ledger = self.standardize_ledger(
                pd.concat([self._ledger, target_df]).drop_duplicates()
            )
```

pyledger/__init__.py
AlexTheWizardL: Correct

tests/__init__.py
AlexTheWizardL: This makes imports clear in the `pyledger/__init__.py` file

pyledger/memory_ledger.py
```diff
@@ -0,0 +1,227 @@
...
+
+        Args:
+            base_currency (str): The base currency for the system. Defaults to "USD".
+        """
+        settings = {
+            "precision": self.PRECISION,
+            "base_currency": base_currency,
+        }
+        super().__init__(
+            settings=settings,
```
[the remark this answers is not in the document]
AlexTheWizardL: Here it is: #32

tests/base_account_chart.py
```diff
@@ -0,0 +1,171 @@
...
+        target_df.loc[target_df["account"] == 2, "text"] = "New_Test_Text"
+        ledger.mirror_account_chart(target_df, delete=True)
+        mirrored_df = ledger.account_chart()
+        m = target_df.merge(mirrored_df, how="outer", indicator=True)
+        assert (m["_merge"] == "both").all(), "Some target accounts were not mirrored"
+
+        ledger.mirror_account_chart(initial_accounts, delete=True)
+        mirrored_df = ledger.account_chart()
+        m = initial_accounts.merge(mirrored_df, how="outer", indicator=True)
+        assert (m["_merge"] == "both").all(), "Some target accounts were not mirrored"
```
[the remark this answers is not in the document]
AlexTheWizardL: Fully agree. That was a context from the `CashCtrlLedger`

tests/base_test_clear.py
```diff
@@ -0,0 +1,27 @@
...
+
+class BaseTestClear(ABC):
+
+    @pytest.fixture()
+    def restore_initial_state(self, ledger, tmp_path):
+        ledger.dump_to_zip(tmp_path / "initial_ledger.zip")
+
+        yield
+
+        ledger.restore_from_zip(tmp_path / "initial_ledger.zip")
```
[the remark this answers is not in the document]
AlexTheWizardL: I haven`t updated old approach

tests/base_test_dump_restore.py
AlexTheWizardL: I haven`t updated old approach

pyledger/memory_ledger.py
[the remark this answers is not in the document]
lasuk: Thanks.

lasuk (commented):
Thanks for implementing the changes. 
Two conversations are left open. 
Back to you @AlexTheWizardL.

tests/test_account_charts.py
lasuk, line 1: Let's rename `test_account_charts.py` to `test_accounts.py`, in line with the future name change from 'account_chart' to 'accounts' proposed in #30.

lasuk:
Thanks for the final changes. Ready to merge ...

It's a pleasure to see our how concept implemented. I find it quite convincing!

PR #34, opened 2024-09-24
Feat: move mirroring methods to LedgerEngine class

AlexTheWizardL:


pyledger/ledger_engine.py
```diff
@@ -445,24 +480,64 @@ def modify_account(self, account: int, new_data: dict) -> None:
         """
 
     @abstractmethod
-    def delete_account(self, account: int, allow_missing: bool = False) -> None:
+    def delete_account(self, accounts: List[int] = [], allow_missing: bool = False) -> None:
         """Removes an account from the account chart.
 
         Args:
-            account (int): The account to be removed.
+            accounts (Union[int, List[int]]): The account number(s) to be deleted.
```
lasuk, line 487: There is a mismatch between type hint and doc string for `accounts`.

lasuk, line 487: Here's a suggestion to clarify the interface: 
- rename `delete_account` to `delete_accounts`. 
- `accounts` is required to be a list, we do not accept singleton as argument.
- Pro: the interface is unambiguous.
- Con: Users have to add brackets when deleting a single account: `delete_accounts([id])` instead of `delete_account(id)`.

The interface would be:
```
def delete_accounts(self, accounts: List[int] = [], allow_missing: bool = False) -> None:
```

rather than:
```
def delete_account(self, accounts: int |  List[int] = [], allow_missing: bool = False) -> None:
```

What do you think?

(applies to all delete_... methods)

pyledger/ledger_engine.py
```diff
@@ -314,15 +314,50 @@ def delete_vat_code(self, code: str, allow_missing: bool = False) -> None:
...
+            delete (bool, optional): If True, deletes VAT codes on the remote account
+                                     that are not present in target_state.
         """
+        target_df = self.standardize_vat_codes(target).reset_index()
+        current_state = self.vat_codes().reset_index()
+
+        # Delete superfluous VAT codes on remote
+        if delete:
+            for idx in set(current_state["id"]).difference(set(target_df["id"])):
+                self.delete_vat_code(code=idx)
```
lasuk, line 333: If delete methods always accept a list, we do not need to loop through vat_codes.

pyledger/ledger_engine.py
```diff
@@ -445,24 +480,64 @@ def modify_account(self, account: int, new_data: dict) -> None:
...
+        if target is not None:
+            target = target.copy()
+        target_df = self.standardize_account_chart(target).reset_index()
+        current_state = self.account_chart().reset_index()
+
+        # Delete superfluous accounts on remote
+        if delete:
+            self.delete_accounts(
+                set(current_state["account"]).difference(set(target_df["account"]))
+            )
```
lasuk, line 510: Marginally better readable:
```python
to_delete = set(current_state["account"]).difference(target_df["account"])
self.delete_accounts(to_delete)
```

Rather than:
```python
self.delete_accounts(
    set(current_state["account"]).difference(set(target_df["account"]))
)
```

pyledger/ledger_engine.py
```diff
@@ -774,22 +849,92 @@ def modify_ledger_entry(self, id: str, new_data: dict) -> None:
...
-        """Aligns ledger entries with a desired target state.
+        """Mirrors the ledger data to the remote CashCtrl instance.
 
         Args:
-            target (pd.DataFrame): DataFrame with ledger entries in the pyledger format.
-            delete (bool, optional): If True, deletes existing ledger that are not
-                                     present in the target data.
+            target (pd.DataFrame): DataFrame containing ledger data in pyledger format.
+            delete (bool, optional): If True, deletes ledger entries on the remote that are
+                                     not present in the target DataFrame.
```
lasuk, line 865: Please document return value,

lasuk, line 865: I like the idea of returning statistics of number of records added, modified, deleted, etc. 
Can we apply the same principle to all mirror_... methods?

pyledger/memory_ledger.py
```diff
@@ -132,24 +119,18 @@ def modify_account(
...
-            raise KeyError(f"Account '{account}' not found in the account chart.")
-        self._account_chart = self._account_chart[self._account_chart["account"] != account]
-
-    def mirror_account_chart(self, target: pd.DataFrame, delete: bool = False):
-        target_df = self.standardize_account_chart(target)
+    def delete_accounts(
+        self, accounts: Union[int, List[int]] = [], allow_missing: bool = False
+    ) -> None:
+        if isinstance(accounts, int):
+            accounts = [accounts]
```
lasuk, line 126: `isinstance(..., int)` is not sufficient to check that account is singular.
account could also be a float or str.

(Only relevant if we decide to not force delete_methods to take a list input. Applies to all delete_.. methods)

lasuk (commented):
Thanks for moving mirror functions to the base class. 
It is a pleasure to see the flexibility of out class structure.

I left a series of comments. Back to you, @AlexTheWizardL.

PR #41, opened 2024-10-11
Feat: create text ledger

AlexTheWizardL:
### This PR introduces the following updates:

1. **Tests for `write_fixed_width_csv`**:
   - Comprehensive test suite covering different scenarios, including handling NA values, fixed-width formatting, and varying column numbers.

2. **`TextLedger` Implementation**:
   - Implemented full functionality for the `ledger` entity, including reading, saving, and managing ledger entries with caching support.
   - Mocked implementations for other entities (`accounts`, `vat_codes`, etc.) based on logic from `MemoryLedger` for future extension.

3. **Test Suite for `TextLedger`**:
   - Tests to verify the caching mechanism, ledger folder creation, and cleanup of unused files/directories.
   - Tests for behavior when reading from nested ledger folders and ensuring paths are correctly stored and managed.

4. **Separate Test File for `save_files` Implementation**:
   - Created a dedicated test file focusing on the `save_files` method.
   - Tests cover various scenarios such as grouping entries by `file_path`, ensuring directories are maintained, and verifying that files/directories not referenced in the DataFrame are correctly removed.

This PR ensures the `ledger` entity's full functionality while establishing robust testing for file handling methods and laying the groundwork for extending `TextLedger` to other entities.


pyledger/text_ledger.py
```diff
@@ -1,251 +1,358 @@
...
+        self.root_path = root_path
+        self._reporting_currency = reporting_currency
+        self._cache_timeout = cache_timeout
+        self._ledger_cache = None
+        self._ledger_cache_time = None
+        self._logger = logging.getLogger(__name__)
+        self._ledger = self.ledger()
+        self._prices = self.standardize_prices(None)
+        self._tax_codes = self.standardize_tax_codes(None)
+        self._accounts = self.standardize_accounts(None)
```
lasuk, line 38: Why do you set empty prices, tax_codes and accounts, but read ledger entries from file?

lasuk (commented):
Thanks for implementing and testing `write_fixed_width_csv`!

I have completed the review for `write_fixed_width_csv` for now and am posting review comments so that you can address them while I continue to review the TextLedger functionality.

pyledger/tests/test_helpers.py
```diff
@@ -0,0 +1,154 @@
...
+        'ShortCol': [1, 2, 3],
+        'LongerCol': ['a', 'bb', 'ccc'],
+        'FinalCol': [10, 20, 30]
+    })
+    output = write_fixed_width_csv(df, n=2)
+    expected_output = ('ShortCol, LongerCol, FinalCol\n'
+                       '       1,         a, 10\n'
+                       '       2,        bb, 20\n'
+                       '       3,       ccc, 30\n')
+    assert output == expected_output, "Failed when n applies to all columns except the last"
```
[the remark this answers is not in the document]
AlexTheWizardL: Yeah, this test seems redundant if we also have `test_fixed_width_columns_partial`

pyledger/tests/test_text_ledger_ledger.py
```diff
@@ -0,0 +1,128 @@
...
+    ledger/level1/ledger2.csv:4,        2024-05-24,   9991,        ,     USD, -100.00,        -88.88,   OutRed,   pytest collective txn 1 - line 1,   /subdir/file2.txt
+    ledger/level1/ledger2.csv:4,        2024-05-24,   9991,        ,     USD,    1.00,          0.89,   OutRed,   pytest collective txn 1 - line 1,   /subdir/file2.txt
+    ledger/level1/ledger2.csv:4,        2024-05-24,   9991,        ,     USD,   99.00,         87.99,   OutRed,   pytest collective txn 1 - line 1,
+    ledger/level1/level2/ledger3.csv:5, 2024-05-24,   9992,    9995,     CHF,  100.00,              ,   OutRed,   pytest single transaction 1,        /file1.txt
+    ledger/level1/level2/ledger3.csv:6, 2024-05-24,   9991,        ,     USD, -100.00,        -88.88,   OutRed,   pytest collective txn 1 - line 1,   /subdir/file2.txt
+    ledger/level1/level2/ledger3.csv:6, 2024-05-24,   9991,        ,     USD,    1.00,          0.89,   OutRed,   pytest collective txn 1 - line 1,   /subdir/file2.txt
+    ledger/level1/level2/ledger3.csv:6, 2024-05-24,   9991,        ,     USD,   99.00,         87.99,   OutRed,   pytest collective txn 1 - line 1,
+"""
+LEDGER_FOR_SAVE_CSV = """
```
lasuk, line 28: Do we need  to write out `LEDGER_FOR_SAVE_CSV`? Could the content not be easily derived from `LEDGER_CSV`?

pyledger/tests/test_text_ledger_ledger.py
```diff
@@ -0,0 +1,128 @@
...
+    ledger/level1/ledger2.csv:6,        2024-05-24,   9991,        ,     USD,    1.00,          0.89,   OutRed,   pytest collective txn 1 - line 1,   /subdir/file2.txt
+    ledger/level1/ledger2.csv:6,        2024-05-24,   9991,        ,     USD,   99.00,         87.99,   OutRed,   pytest collective txn 1 - line 1,
+"""
+# flake8: enable
+
+class TestLedger(BaseTestLedger):
+
+    @pytest.fixture
+    def ledger(self, tmp_path):
+        return TextLedger(tmp_path / "system")
```
lasuk, line 49: Why do we need a "system" directory?

pyledger/text_ledger.py
```diff
@@ -1,251 +1,358 @@
...
-        Args:
-            settings (str): Path to the JSON settings file.
-            accounts (str): Path to the CSV accounts file.
-            ledger (str): Path to the directory containing ledger CSV files.
-            prices (str, optional): Path to the directory containing price CSV files.
-            tax_codes (str, optional): Path to the CSV tax codes file.
-            revaluations (str, optional): Path to the CSV revaluations file.
+        root_path: Path = Path.cwd(),
+        cache_timeout: int = 300,
+        reporting_currency: str = "USD",
```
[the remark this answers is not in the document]
AlexTheWizardL: This is only a placeholder for now and will be removed

pyledger/tests/test_text_ledger_ledger.py
```diff
@@ -0,0 +1,128 @@
...
+        df.to_csv(ledger_dir / "ledger1.csv", index=False)
+        df.to_csv(ledger_dir / "level1/ledger2.csv", index=False)
+        df.to_csv(ledger_dir / "level1/level2/ledger3.csv", index=False)
+
+        ledger = TextLedger(root_path=tmp_path)
+        real_ledger = ledger.ledger()
+        expected_ledger = ledger.standardize_ledger(pd.read_csv(
+            StringIO(LEDGER_CSV), skipinitialspace=True
+        ))
+        assert_frame_equal(real_ledger, expected_ledger, ignore_row_order=True)
```
lasuk, line 81: I do not understand the motivation for this test.

pyledger/text_ledger.py
AlexTheWizardL: This is also a placeholder for now and will be removed with the next implementation

pyledger/tests/test_text_ledger_ledger.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def test_ledger_read_empty_files(self, tmp_path):
```
lasuk, line 83: The purpose of this test was not immediately clear to me.
A one-line doc string explaining the purpose would facilitate quick understanding of the code.

pyledger/tests/test_text_ledger_ledger.py
```diff
@@ -0,0 +1,128 @@
...
+        (ledger_dir / "ledger1.csv").touch()
+        (ledger_dir / "level1/ledger2.csv").touch()
+        (ledger_dir / "level1/level2/ledger3.csv").touch()
+
+        ledger = TextLedger(root_path=tmp_path)
+        output = ledger.ledger()
+        expected_ledger = ledger.standardize_ledger(None)
+        assert_frame_equal(output, expected_ledger, ignore_row_order=True)
+
+    def test_ledger_read_no_ledger_folder(self, tmp_path):
```
lasuk, line 97: The purpose of this test was not immediately clear to me.
Either a one-line doc string or more explicit method name would be helpful, e.g. `test_read_ledger_from_empty_folder()`

pyledger/tests/test_text_ledger_ledger.py
```diff
@@ -0,0 +1,128 @@
...
+        ledger = TextLedger(root_path=tmp_path)
+        output = ledger.ledger()
+        expected_ledger = ledger.standardize_ledger(None)
+        assert_frame_equal(output, expected_ledger, ignore_row_order=True)
+
+    def test_ledger_creates_ledger_folder_on_init(self, tmp_path):
+        TextLedger(tmp_path)
+        ledger_folder = tmp_path / "ledger"
+        assert ledger_folder.exists(), "The 'ledger' folder should be created."
+        assert ledger_folder.is_dir(), "The 'ledger' path should be a directory."
```
lasuk, line 107: Why does TextLedger create an empty ledger folder on init? I would expect the file system to remain unchanged until we call a mutator method.

pyledger/text_ledger.py
```diff
@@ -1,251 +1,358 @@
...
+    ):
+        """Initializes the TextLedgerClient with a root path for file storage.
+        If no root path is provided, defaults to the current working directory.
         """
```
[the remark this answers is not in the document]
AlexTheWizardL: Just forgot to do so

pyledger/tests/test_text_ledger_ledger.py
```diff
@@ -0,0 +1,128 @@
...
+            "ledger/level1/ledger2.csv"
+        ], name="file_path")
+
+        ledger_entries = pd.read_csv(StringIO(LEDGER_FOR_SAVE_CSV), skipinitialspace=True)
+        expected_ledger = ledger.standardize_ledger(ledger_entries.copy())
+        expected_ledger.drop(columns="id", inplace=True)
+        expected_ledger["date"] = dates
+        expected_ledger["file_path"] = file_paths
+        output = ledger._ledger_for_save(ledger_entries)
+        assert_frame_equal(output, expected_ledger, ignore_row_order=True)
```
lasuk, line 128: I do not understand the purpose of this test.

pyledger/constants.py
```diff
@@ -58,6 +58,10 @@
 """
 LEDGER_SCHEMA = pd.read_csv(StringIO(LEDGER_SCHEMA_CSV), skipinitialspace=True)
 
+FIXED_LEDGER_COLUMNS = [
+    col for col in LEDGER_SCHEMA["column"] if col not in ["description", "document", "id"]
```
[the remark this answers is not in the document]
AlexTheWizardL: It is not used while saving to the CSV file, that's why I don't count it

pyledger/text_ledger.py
lasuk: Ok, you should indicate this with a TODO. Please comment on what exactly you consider to be a placeholder;  the planned outcome.

pyledger/text_ledger.py
```diff
@@ -1,251 +1,358 @@
...
-            accounts=_read_csv_or_none(accounts),
-            ledger=self.read_ledger(ledger),
-            prices=self.read_prices(prices),
-            tax_codes=_read_csv_or_none(tax_codes),
-            revaluations=_read_csv_or_none(revaluations),
-        )
+        self.root_path = root_path
+        self._reporting_currency = reporting_currency
+        self._cache_timeout = cache_timeout
+        self._ledger_cache = None
```
lasuk, line 46: Why do we need a `_ledger_cache` variable? Would `_ledger` not suffice?

pyledger/text_ledger.py
lasuk, line 41: You could initialise local variables `tax_codes`, `accounts`, `ledger`, etc. here to let readers know of their existence and type.

pyledger/text_ledger.py
```diff
@@ -1,251 +1,358 @@
...
+        self._tax_codes = self.standardize_tax_codes(None)
+        self._accounts = self.standardize_accounts(None)
+
+    def ledger(self, short_names: bool = False) -> pd.DataFrame:
+        if self._ledger_cache is not None and not self._is_expired(self._ledger_cache_time):
+            return self._ledger_cache
+
+        ledger = self.standardize_ledger(None)
+        ledger_folder = self.root_path / "ledger"
+        ledger_folder.mkdir(parents=True, exist_ok=True)
```
lasuk, line 56: Why create a folder in a read method? Just exit if the folder does not exist.

pyledger/text_ledger.py
```diff
@@ -1,251 +1,358 @@
...
+        for file in ledger_files:
+            try:
+                path = file.relative_to(self.root_path)
+                df = pd.read_csv(file, skipinitialspace=True)
+                if not df.empty:
+                    df["file_path"] = str(path)
+                    if ledger.empty:
+                        ledger = df.copy()
+                    else:
+                        ledger = pd.concat([ledger, df], ignore_index=True)
```
lasuk, line 67: Serial concatenation in a loop is inefficient.
Instead, you should append the data frames to a list and concatenate the resulting list.

```
result = []
for file in ledger_files:
    try:
         df = pd.read_csv(file, skipinitialspace=True)
         if not df.empty:
             df["file_path"] = str(file.relative_to(self.root_path))
             result.append(self.standardize_ledger_columns(df))
    except Exception as e:
        self._logger.warning(f"Skipping {path}: {e}")
result = pd.concat(result, ignore_index=True)
```

pyledger/text_ledger.py
```diff
@@ -1,251 +1,358 @@
...
+            # TODO: remove this code block when old system will be migrated
+            if short_names:
+                LEDGER_COLUMN_SHORTCUTS = {
+                    "cur": "currency",
+                    "vat": "tax_code",
+                    "target": "target_balance",
+                    "base_amount": "report_amount",
+                    "counter": "contra",
+                }
+                ledger = ledger.rename(columns=LEDGER_COLUMN_SHORTCUTS)
```
lasuk, line 87: Column renaming is best executed separately for each fie when reading files in the loop.
This allows to combine files with different naming patterns.

lasuk (commented):
Posting the next batch of comments while continuing to review. 
I've reviewed the code up to and including reading ledger files.

pyledger/text_ledger.py
```diff
@@ -1,251 +1,377 @@
...
-        df: pd.DataFrame,
-        path: str,
-        drop_unused_columns: bool = False,
-        digits: int | None = None,
-    ) -> None:
-        """Writes a ledger DataFrame to a CSV file.
-        CSV files are formatted with fixed column width and standard column
-        order to improve human readability and traceability with a version
-        control system such as git.
+    def save_files(self, df: pd.DataFrame, folder: str, fixed_n: int = None) -> None:
```
lasuk, line 191: `save_files` should in my opinion a helper function without class reference.
You can either decorate it as static method or move it to the helpers module.

Existing references to self are not needed:
- Log entries can be dropped. I do not know of a use case where they are beneficial.
- self.root_path is not needed. This path should be part of the 'root' argument.

pyledger/text_ledger.py
```diff
@@ -1,251 +1,377 @@
...
 
-        # Default column order
-        cols = (
-            [col for col in LEDGER_COLUMN_SEQUENCE if col in df.columns]
-            + [col for col in df.columns if col not in LEDGER_COLUMN_SEQUENCE]
+        if fixed_n is None:
+            fixed_n = len(df.columns)
+        default_folder = self.root_path / folder
+        default_folder.mkdir(parents=True, exist_ok=True)
+        if "file_path" not in df.columns:
```
lasuk, line 209: - Please rename the "file_path" column to "__path__" to reduce the chance of an accidental confusion with a data frame column.
- Let's require the presence of the "__path__" column, rather than proposing a default value. After all, this is a pretty peculiar method that will never be used in a general context. The method works in tandem with `_ledger_for_save`, which is responsible for adding the "__path__" column.

pyledger/text_ledger.py
lasuk, line 191: The method seems overly complicated. Here's a sketch for a simpler version:
```python
def save_files(df: pd.DataFrame, root: Path | str, n_fixed: int = None):
    """
    Save a data frame to multiple csv files specified in the '__csv_path__' column.

    Ensures that csv files in a `root` folder contain the data in `df`.
    1. Saves df to multiple csv files specified by df["__csv_path__"].
    2. Removes unneeded csv files.
    3. Deletes any empty directories.

    Args:
        df (pd.DataFrame): The DataFrame containing the data to be saved
                            and a '__csv_path__' column.
        root (str): The root folder in which csv are stored.
        n_fixed (int, optional): Number of columns from the left to write with fixed column width.
    """
    if n_fixed is None:
        n_fixed = len(df.columns)
    if "__csv_path__" not in df.columns:
        raise Exception("...")

    # Ensure root exists and is a directory
    root = Path(root).expanduser()
    root.mkdir(parents=True, exist_ok=True)

    # Delete unneeded csv files
    for file in root.rglob("*.csv"):
        if not file in df['__csv_path__']:
            file.unlink()
    # As an alternative, you can also use set operations:
    # for file in set(root.rglob("*.csv")).difference(df['__csv_path__']):
    #    file.unlink()

    # Save DataFrame entries to their respective file paths
    for path, group in df.groupby("__csv_path__"):
        full_path = Path(root) / path
        full_path.parent.mkdir(parents=True, exist_ok=True)
        write_fixed_width_csv(group, full_path, n=n_fixed)

    # Delete empty directories
    # Ask ChatGPT for help deleting all empty directories in root.
```

lasuk (commented):
Here's a few additional comments. I need to go now, will continue tomorrow.

pyledger/tests/test_text_ledger_ledger.py
AlexTheWizardL: Actually no, good catch, thanks!

pyledger/tests/test_text_ledger_ledger.py
AlexTheWizardL: It was just for my own purposes, I can simplify and remove it.

pyledger/tests/test_text_ledger_ledger.py
AlexTheWizardL: I`ll try to add a short docstring and comments to clarify test steps.

pyledger/tests/test_text_ledger_ledger.py
AlexTheWizardL: I`ll add docstring for clarification.

pyledger/tests/test_text_ledger_ledger.py
AlexTheWizardL: I`ll add docstring for clarification.

pyledger/tests/test_text_ledger_ledger.py
AlexTheWizardL: Make sense to me, I would change this.

pyledger/tests/test_text_ledger_ledger.py
AlexTheWizardL: I`ll try to add a short docstring and comments to clarify test steps.

pyledger/text_ledger.py
```diff
@@ -1,251 +1,377 @@
...
-                df_list.append(df)
-            except Exception as e:
-                self._logger.warning(f"Skip {short_file}: {e}")
-        if len(df_list) > 0:
-            result = pd.concat(df_list, ignore_index=True)
-        else:
-            # Empty DataFrame with identical structure
-            cols = LEDGER_SCHEMA["column"]
-            result = self.standardize_ledger(pd.DataFrame(columns=cols))
+    def add_ledger_entry(self, entry: pd.DataFrame) -> str:
```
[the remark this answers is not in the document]
AlexTheWizardL: In the case of TextLedger - ID will be ignored and assigned new one
In the case of MemoryLedger, I`ll make an update to ensure it isn't exists

lasuk:
@AlexTheWizardL: I find the proposed text ledger implementation too complex, with some parts being convoluted and more invasive than necessary. I’ve committed a sketch for a simplified version  in 2fdcb07c7.

**Main contributions:**

1. `write_ledger_file(df, file)` saves a single ledger file in a canonical format.
    - This method defines the standard text format for ledger files, following a hard-coded, deterministic format without customization.
    - It is a static method, meaning it can be called without instantiating a text ledger class. For example, a booking rule can use it to write a ledger file directly from bank transaction entries.

2. `write_ledger_directory(df, root)` saves ledger entries across multiple files in a root directory.
    - Unnecessary CSV files are deleted automatically.
    - Unlike the first implementation, there’s no need to remove empty directories.

3. `[add|modify|delete]_ledger_entry` methods are minimally invasive, affecting only the specific ledger file(s) they modify.
    - This approach is sketched out in `add_ledger_entry` and can be similarly applied to `modify` and `delete` methods.

4. When reading a directory with ledger entries, each file is standardized separately.
    - Any error in reading or standardizing data only affects the specific file with the issue.
    - The `id` field’s numbering starts from zero for each file.

AlexTheWizardL:
### **Section 1: Reading Ledger Entries**

1. **Check Directory**:  
   Verify if the ledger directory exists. If not, return an empty standardized DataFrame. If the path is invalid, raise an exception.

2. **Iterate Through Files**:  
   Use `Path.rglob()` to find all CSV files in the ledger directory.

3. **Read and Process Each File**:  
   - **Standardize the DataFrame**: For each CSV file, read its contents into a DataFrame using `pd.read_csv()`. Rename columns using `LEDGER_COLUMN_SHORTCUTS` to ensure consistency and prepend the file name to the `id` column to ensure unique `id`s across files.
   - **Schema Enforcement**: Call `self.standardize_ledger()` on each DataFrame to ensure it conforms to the ledger schema (e.g., data types, column order). Each file is processed separately before moving on to the next.

4. **Concatenate DataFrames**:  
   Once all files are processed and standardized, concatenate the DataFrames into a single ledger DataFrame.

5. **Cache Expiration Check**:  
   Before reading files, check if the cached ledger has expired. If expired, reload and process the files again. If the cache is valid, return the cached data without reloading.

---

### **Code Implementation for Section 1:**

```python
def ledger(self) -> pd.DataFrame:
    if self._is_expired(self._ledger_time):
        self._ledger = self.read_ledger_files(self.root_path / "ledger")
        self._ledger_time = datetime.now()
    return self._ledger

def read_ledger_files(self, root: Path) -> pd.DataFrame:
    # Check if directory exists
    if not root.exists():
        return self.standardize_ledger(None)
    if not root.is_dir():
        raise NotADirectoryError(f"Ledger root folder is not a directory: {root}")
    
    ledger = []
    
    # Iterate through all CSV files
    for file in root.rglob("*.csv"):
        try:
            # Read file contents into a DataFrame
            df = pd.read_csv(file, skipinitialspace=True)
            # Standardize column names
            df = df.rename(columns=LEDGER_COLUMN_SHORTCUTS)
            # Ensure the DataFrame conforms to the schema
            df = self.standardize_ledger(df)
            
            # Handle ID uniqueness by appending file name to 'id'
            if not df.empty:
                df["id"] = str(file.relative_to(root)) + ":" + df["id"]
            ledger.append(df)
        except Exception as e:
            self._logger.warning(f"Skipping file {file}: {e}")
    
    # Concatenate DataFrames
    if ledger:
        result = pd.concat(ledger, ignore_index=True)
    else:
        result = pd.DataFrame()
    
    # Standardize the final result
    return self.standardize_ledger(result)

def _is_expired(self, cache_time: datetime) -> bool:
    return (datetime.now() - cache_time) > timedelta(seconds=self._cache_timeout)
```

---

### **Section 2: Saving Ledger Entries**

1. **Add/Modify/Delete Methods Trigger Changes**:  
   Whenever a ledger file is modified, these methods invoke `write_ledger_directory()` to handle saving the changes. Only the relevant portion of the DataFrame is passed to ensure efficient processing.

2. **Standardize Data and Define `__csv_path__`:**  
   Inside `write_ledger_directory()`, the DataFrame is standardized by calling `self.standardize_ledger()` to ensure schema compliance (e.g., correct column types and order). A `__csv_path__` column is created, derived from the `id` column, to specify the file path where each ledger entry will be saved.

3. **File Preparation with `write_ledger_file()`**:  
   - `write_ledger_file()` takes each grouped DataFrame and prepares it for saving:
     - Enforces the correct data types and ensures the column and row order matches the ledger schema.
     - Drops unnecessary columns like `id`.
     - Formats the `date` column correctly.
   - This method is responsible for the detailed preparation of the DataFrame before it’s saved as a fixed-width CSV file.

4. **Save Files and Clean Up with `save_files()`**:  
   After preparing the DataFrame, it is passed to the `save_files()` function. This function:
   - Groups ledger entries by the `__csv_path__` column, ensuring that entries for the same file are processed together.
   - Calls the `write_ledger_file()` method for each group to handle detailed DataFrame preparation (e.g., enforcing schema, dropping unnecessary columns, formatting the `date` column).
   - Saves the prepared entries to the correct files, ensuring they are placed in the appropriate nested folder structure.
   - Removes any files in the ledger directory that are no longer referenced by the DataFrame, keeping the directory clean and up-to-date.

---

### **Code Implementation for Section 2:**

```python
def write_ledger_directory(self, df: pd.DataFrame) -> pd.DataFrame:
    """Save ledger entries to multiple fixed-width CSV files.

    Args:
        df (pd.DataFrame): The ledger DataFrame to prepare.

    Returns:
        pd.DataFrame: The formatted ledger DataFrame ready for saving.
    """
    df = self.standardize_ledger(df)
    df["__csv_path__"] = self.csv_path(df["id"])
    save_files(df, root=self.root_path / "ledger", func=self.write_ledger_file)
    return df

@classmethod
def write_ledger_file(cls, df: pd.DataFrame, file: Path):
    """Save ledger entries to a fixed-width CSV file.

    Args:
        df (pd.DataFrame): Ledger entries to save.
        file (Path): The file path where the entries will be saved.
    """
    df = enforce_schema(df, LEDGER_SCHEMA, sort_columns=True, keep_extra_columns=True)
    df["date"] = df["date"].where(~df.duplicated(subset="id"), None)
    df = df.drop(columns=["id"])  # Drop 'id' column and potentially other NA columns
    save_fixed_width_csv(df, file=file)

def save_files(df: pd.DataFrame, root: Path, func) -> None:
    """Save a DataFrame to multiple CSV files specified in the '__csv_path__' column.

    Args:
        df (pd.DataFrame): The DataFrame containing ledger data and file paths.
        root (Path): The root folder where the CSV files will be saved.
        func: A function that takes a DataFrame and a file path to save the file.
    """
    if "__csv_path__" not in df.columns:
        raise ValueError("The DataFrame must contain a '__csv_path__' column.")

    root = Path(root).expanduser()
    root.mkdir(parents=True, exist_ok=True)

    # Delete unneeded files
    current_files = set(root.rglob("*.csv"))
    referenced_files = set(root / path for path in df["__csv_path__"].unique())
    for file in current_files - referenced_files:
        file.unlink()

    # Save DataFrame entries to their respective file paths
    for path, group in df.groupby("__csv_path__"):
        if path:
            full_path = root / path
            full_path.parent.mkdir(parents=True, exist_ok=True)
            func(group.drop(columns="__csv_path__"), full_path)
```

---

### Small Question:

In my understanding, the `write_ledger_directory()` method may not add significant value. Removing it and handling standardization and file path definition directly within the Add/Modify/Delete methods could simplify the code and improve clarity. Do you think removing it would make the code easier to follow, or is there a benefit to keeping it that I'm overlooking?

AlexTheWizardL:
### **1. Testing the Reading Process**

#### **Goal**: 
Ensure that all ledger files are correctly read, processed, standardized, and concatenated, while handling potential errors gracefully.

#### **Key Scenarios to Test**:

1. **Directory Exists**:  
   Test that the method correctly reads the files from a directory and returns a standardized DataFrame.

2. **Directory Does Not Exist**:  
   Test that if the directory doesn't exist, an empty DataFrame is returned.

3. **Invalid Directory Path**:  
   Test that if the provided path is not a directory, a `NotADirectoryError` is raised.

4. **Multiple CSV Files**:  
   Test that multiple CSV files are read, standardized, and concatenated correctly into a single DataFrame.

5. **Handling Ill-Formatted Files**:  
   Test how the system behaves if one of the CSV files is ill-formatted (e.g., missing columns, incorrect delimiters) and ensure that such files are skipped without crashing.

6. **Cache Expiration**:  
   Test if the cached ledger is reloaded correctly when the cache has expired.

---

### **2. Testing the Saving Process**

#### **Goal**: 
Ensure that ledger entries are correctly standardized, saved to the appropriate file paths, and any unneeded files are removed.

#### **Key Scenarios to Test**:

1. **Saving New Ledger Entries**:  
   Test that new ledger entries are correctly saved to the appropriate CSV files.

2. **Schema Enforcement**:  
   Test that the saved ledger files are standardized correctly (i.e., correct data types, column order).

3. **Unreferenced File Cleanup**:  
   Test that files no longer referenced in the ledger are removed after saving.

---

### **3. Testing the `save_files()` Method**

#### **Goal**:  
Ensure that `save_files()` correctly saves data to the appropriate files and deletes unreferenced files.

#### **Key Scenarios to Test**:

1. **Saving Data to the Correct File Paths**:  
   Test that the DataFrame entries are saved to the CSV files specified by the `__csv_path__` column.

2. **Handling Multiple File Paths**:  
   Test that `save_files()` correctly groups DataFrame entries by the `__csv_path__` column and saves each group to the appropriate file.

3. **Unreferenced File Cleanup**:  
   Test that files in the directory that are no longer referenced by the DataFrame are deleted.

4. **Handling Empty Files**:  
   Test that empty files are deleted or not created when there are no entries to be saved for a particular `__csv_path__`.

pyledger/text_ledger.py
```diff
@@ -1,251 +1,344 @@
...
+    def modify_ledger_entry(self, entry: pd.DataFrame) -> None:
+        entry = self.standardize_ledger(entry)
+        if entry["id"].nunique() != 1:
+            raise ValueError(
+                "Id needs to be unique and present in all rows of a collective booking."
+            )
+
+        ledger = self.ledger()
+        path = self.csv_path(entry["id"]).iloc[0]
+        df_same_file = ledger[ledger["id"].str.startswith(path)]
```
lasuk, line 186: Clearer:
```
df_same_file = ledger[self.csv_path(ledger["id"]) == path]
```

rather than:
```
df_same_file = ledger[ledger["id"].str.startswith(path)]
```

pyledger/text_ledger.py
```diff
@@ -1,251 +1,344 @@
...
+
+    def delete_ledger_entries(self, ids: List[str], allow_missing: bool = False) -> None:
+        ledger = self.ledger()
+        if not allow_missing:
+            missing_ids = set(ids) - set(ledger["id"])
+            if missing_ids:
+                raise KeyError(f"Ledger entries with ids '{', '.join(missing_ids)}' not found.")
+
+        df = ledger[~ledger["id"].isin(ids)]
+        paths_to_update = ledger["id"].str.split(":").str[0].unique()
```
lasuk, line 202: For consitency, Please always use our method to extract the path
```
paths_to_update = self.csv_path(ledger["id"]).unique()
```

rather than:
```
paths_to_update = ledger["id"].str.split(":").str[0].unique()
```

pyledger/text_ledger.py
```diff
@@ -1,251 +1,344 @@
...
+        for path in paths_to_update:
+            df_same_file = df[df["id"].str.startswith(path)]
```
lasuk, line 204: Clearer:
```
df_same_file = df[self.csv_path(df["id"]) == path]
```

rather than:
```
df_same_file = df[df["id"].str.startswith(path)]
```

pyledger/text_ledger.py
```diff
@@ -1,251 +1,344 @@
...
+            self.write_ledger_file(df_same_file, self.root_path / path)
```
lasuk, line 205: path is specified relative to `root/ledger` directory.

Pleas use:
```
self.write_ledger_file(df_same_file, self.root_path / "ledger" / path)
```

rather than:
```
self.write_ledger_file(df_same_file, self.root_path / path)
```

lasuk, line 205: We decided to erase file that do not have any ledger entries any more.

lasuk (commented):
Thanks for the updated PR.

@AlexTheWizardL: I have reviewed everything except the tests for now. Need to leave now, will continue tomorrow.

pyledger/text_ledger.py
```diff
@@ -1,251 +1,344 @@
...
-        errors.
-
-        Args:
-            path (str, optional): Path to the directory containing price CSV files.
+        super().__init__(settings=SETTINGS)
+        self.root_path = Path(root_path).expanduser()
+        self._reporting_currency = reporting_currency
+        self._cache_timeout = cache_timeout
+        self._ledger_time = None
+        self._ledger = self.ledger()
```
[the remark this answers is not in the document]
AlexTheWizardL: I thought it would be helpful to read the system on the init.

pyledger/text_ledger.py
```diff
@@ -1,251 +1,344 @@
...
-            result = pd.concat(df_list, ignore_index=True)
-        else:
-            # Empty DataFrame with identical structure
-            cols = LEDGER_SCHEMA["column"]
-            result = self.standardize_ledger(pd.DataFrame(columns=cols))
+                self._logger.warning(f"Skipping {file} file: {e}")
+        result = None
+        if ledger:
+            result = pd.concat(ledger, ignore_index=True)
+            result = enforce_schema(result, LEDGER_SCHEMA, sort_columns=True)
```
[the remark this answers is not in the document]
AlexTheWizardL: I can not reply to the previous comment regarding the __init__ method, so replying here:

```python
    _ledger = self.standardize_ledger(None)
    _prices = self.standardize_prices(None)
    _tax_codes = self.standardize_tax_codes(None)
    _accounts = self.standardize_accounts(None)
```

is not possible to call, these assignments reference self, but class-level variables do not have access to instance attributes such as self. This will cause an error since self is not defined at the class level.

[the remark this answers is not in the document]
AlexTheWizardL: > Do we need this line? `enforce_schema` is agin called inside `standardize_ledger` below.

Yes, with this one I enforce column order

lasuk:
I found  some argument for why we should not change the sequence of ledger entries when saving a file.
- Changing the row sequence will alter the 'id' of existing ledger entries.
- When adding a ledger entry, our logic for finding the id only works if the new entry is added last (otherwise existing entries change the id). We should therefore not order by data since a new entry with an earlier date than existing entries would not be added last.

lasuk:
Along the same line of argumentation, please change 'modify_ledger_entry' such that the newly modified entry is added at the same place as the earlier version with the same id. Ordering ledger entries by their id in `write_ledger_entry` should do the trick.

lasuk:
I've added a series of commits to clarify code and doc strings and make error messages more concise.

Also, I've pushe two commits fixing erroneous behaviour:
- c645466 sorts ledger entries on integer rather than string representation of numeric portion of id
- ba69dbd only overwrites affected files during deletion.

AlexTheWizardL:
## `argsort()` (the Pitfall)

```python
import pandas as pd

data = {
    "id": [
        "default.csv:1", "default.csv:2", "default.csv:2", "default.csv:2", 
        "default.csv:3", "default.csv:3", "default.csv:4", "default.csv:5", 
        "default.csv:5", "default.csv:6", "default.csv:6", "default.csv:7", 
        "default.csv:7", "default.csv:8", "default.csv:9", "default.csv:9", 
        "default.csv:9"
    ],
    "value": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17]
}

df = pd.DataFrame(data)
df["id_numeric"] = df["id"].str.split(":").str[1].astype(int)
```

Now, the "value" column looks like this:

```
               id  value  id_numeric
0   default.csv:1      1           1
1   default.csv:2      2           2
2   default.csv:2      3           2
3   default.csv:2      4           2
4   default.csv:3      5           3
5   default.csv:3      6           3
6   default.csv:4      7           4
7   default.csv:5      8           5
8   default.csv:5      9           5
9   default.csv:6     10           6
10  default.csv:6     11           6
11  default.csv:7     12           7
12  default.csv:7     13           7
13  default.csv:8     14           8
14  default.csv:9     15           9
15  default.csv:9     16           9
16  default.csv:9     17           9
```

### Applying `argsort()` (the Pitfall)

When you apply `argsort()` to the `id_numeric` column, the DataFrame will be sorted by `id_numeric`, but the order of rows with the same `id_numeric` value might change:

```python
sorted_indices = df["id_numeric"].argsort()
df_sorted = df.iloc[sorted_indices]
print(df_sorted)
```

The result:

```
               id  value  id_numeric
0   default.csv:1      1           1
1   default.csv:2      2           2
2   default.csv:2      3           2
3   default.csv:2      4           2
4   default.csv:3      5           3
5   default.csv:3      6           3
6   default.csv:4      7           4
8   default.csv:5      9           5
7   default.csv:5      8           5
9   default.csv:6     10           6
10  default.csv:6     11           6
11  default.csv:7     12           7
12  default.csv:7     13           7
13  default.csv:8     14           8
15  default.csv:9     16           9
14  default.csv:9     15           9
16  default.csv:9     17           9
```

### Notice the Reordering

In the output of `argsort()`, observe:

- **For the value `5`**:
  - Index `7` (with value `8`) originally appeared **before** index `8` (with value `9`) in the DataFrame. However, in the `argsort()` result, **index `8` comes before index `7`**.
  
- **For the value `9`**:
  - Index `14` (with value `15`) originally appeared **before** indices `15` and `16` in the original DataFrame. In the `argsort()` result, **index `14` is placed after indices `15` and `16`**.

### Why Does the Order Change?

The change in order for equal values might be due to:

1. **Internal implementation of `argsort()`**:
   By default, pandas’ `argsort()` uses a stable sorting algorithm (`mergesort`), which *should* preserve the order of equal elements. However, in some cases, factors like the data structure might lead to reordering.

2. **Underlying data types or optimizations**:
   String to integer conversion or internal optimizations might also affect how pandas handles sorting, leading to unexpected reordering.

### Solution: Use `kind="mergesort"` parameter.

Here is the link for official [argsort documenttion ](https://numpy.org/doc/2.0/reference/generated/numpy.argsort.html#:~:text=array%20is%20used.-,kind,-%7B%E2%80%98quicksort%E2%80%99%2C%20%E2%80%98mergesort%E2%80%99%2C%20%E2%80%98heapsort)


```python
df = df.iloc[TextLedger.id_from_path(df["id"]).argsort(kind="mergesort")]
```

Now, the output maintains the original order:

```
               id  value  id_numeric
0   default.csv:1      1           1
1   default.csv:2      2           2
2   default.csv:2      3           2
3   default.csv:2      4           2
4   default.csv:3      5           3
5   default.csv:3      6           3
6   default.csv:4      7           4
7   default.csv:5      8           5
8   default.csv:5      9           5
9   default.csv:6     10           6
10  default.csv:6     11           6
11  default.csv:7     12           7
12  default.csv:7     13           7
13  default.csv:8     14           8
14  default.csv:9     15           9
15  default.csv:9     16           9
16  default.csv:9     17           9
```

lasuk:
Thanks a lot for catching this problem, for the fix, and for the clear explanation.

We should remember to always specify a sorting algorithm when sorting is ambiguous (i.e. when the sorted values can contain duplicates). Applies both to [argsort](https://pandas.pydata.org/docs/reference/api/pandas.Series.argsort.html) and to [sort_values](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.sort_values.html).

lasuk:
Merged. Thanks for the great work and the endurance!

PR #51, opened 2024-10-31
Implement formatting with the smallest increment for `TextLedger` ledger entity float values

AlexTheWizardL:
Implement Consistent Floating Point Formatting for "Amount" Columns

To reduce unnecessary git diffs caused by inconsistent floating point representations, this PR introduces a mechanism that standardizes floating point formatting across "amount" columns. The approach identifies the smallest increment used for each currency in the column within a file and applies this precision uniformly across the entire column before saving. This ensures a consistent, minimal format for all entries, improving diff readability and reducing clutter in version control.

### DIFF
--- a/pyledger/standalone_ledger.py
+++ b/pyledger/standalone_ledger.py
@@ -8,7 +8,7 @@ from typing import Dict
 import numpy as np
 import pandas as pd
 from pyledger.decorators import timed_cache
-from .constants import LEDGER_SCHEMA
+from .constants import DEFAULT_ASSETS, LEDGER_SCHEMA
 from .ledger_engine import LedgerEngine
 from consistent_df import enforce_schema
 
@@ -341,7 +341,11 @@ class StandaloneLedger(LedgerEngine):
 
         increment = self._assets_as_dict_of_df.get(ticker)
         if increment is None:
-            raise ValueError(f"No asset definition available for ticker '{ticker}'.")
+            if not DEFAULT_ASSETS["ticker"].unique():
+                raise ValueError("Default assets must contain unique ticker values.")
+            increment = DEFAULT_ASSETS.query(f"ticker == '{ticker}'")
+            if increment.empty:
+                raise ValueError(f"No asset definition available for ticker '{ticker}'.")
 
         mask = increment["date"].isna() | (increment["date"] <= pd.Timestamp(date))
         if not mask.any():
--- a/pyledger/standalone_ledger.py
+++ b/pyledger/standalone_ledger.py
@@ -339,15 +339,19 @@ class StandaloneLedger(LedgerEngine):
         elif not isinstance(date, datetime.date):
             date = pd.to_datetime(date).date()
 
-        increment = self._assets_as_dict_of_df.get(ticker)
-        if increment is None:
-            if not DEFAULT_ASSETS["ticker"].unique():
-                raise ValueError("Default assets must contain unique ticker values.")
-            increment = DEFAULT_ASSETS.query(f"ticker == '{ticker}'")
-            if increment.empty:
+        asset = self._assets_as_dict_of_df.get(ticker)
+        if asset is None:
+            # Asset is not defined by the user, fall back to hard-coded defaults
+            increment = DEFAULT_ASSETS.loc[DEFAULT_ASSETS["ticker"] == ticker, "increment"]
+            if len(increment) < 1:
                 raise ValueError(f"No asset definition available for ticker '{ticker}'.")
-
-        mask = increment["date"].isna() | (increment["date"] <= pd.Timestamp(date))
-        if not mask.any():
-            raise ValueError(f"No asset definition available for '{ticker}' on or before {date}.")
-        return increment.loc[mask[mask].index[-1], "increment"]
+            if len(increment) > 1:
+                raise ValueError(f"Multiple default definitions for asset '{ticker}'.")
+            return increment.item()
+        else:
+            mask = asset["date"].isna() | (asset["date"] <= pd.Timestamp(date))
+            if not mask.any():
+                raise ValueError(
+                    f"No asset definition available for '{ticker}' on or before {date}."
+                )
+            return asset.loc[mask[mask].index[-1], "increment"].item()
--- a/pyledger/text_ledger.py
+++ b/pyledger/text_ledger.py
@@ -1,5 +1,7 @@
 """This module defines TextLedger, extending StandaloneLedger to store data in text files."""
 
+from datetime import datetime
+import math
 import pandas as pd
 import yaml
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


pyledger/text_ledger.py
```diff
@@ -204,6 +204,11 @@ def write_ledger_file(df: pd.DataFrame, file: str) -> pd.DataFrame:
         without a date belong to the transaction that began in the preceding
         row with a date.
 
+        The "amount" and "report_amount" columns are formatted to ensure consistent decimal
+        places across entries, aligning all amounts to the same digit length. This enhances
+        readability and makes it easier to review changes in the CSV file. Missing values
+        are replaced with an empty string, and original significant digits are preserved.
```
lasuk, line 210: Proposal for clarified doc string:

```
"""Saves ledger entries to a fixed-width CSV file.

This function saves ledger entries to a fixed-width CSV file with
consistent formatting for enhanced readability, ideal for version
control systems like Git. Entries are padded with spaces to maintain
consistent column widths. Floating-point numbers in the same column
are right-aligned with a uniform number of decimal places to align
decimal points.

The "amount" and "report_amount" columns are formatted with the
number of decimal places required to accommodate the smallest price
increment of their respective currencies. This ensures concise Git
diffs when new entries with maximum allowed decimal digits are added.

The "id" column is excluded from the saved file. For transactions
that span multiple rows sharing the same "id", the date is recorded
only on the first row. Subsequent rows without a date are considered
part of the transaction that began in the preceding row with a date.
```

lasuk (commented):
Thanks for implementing ledger files with uniform precision.

I left a few comments. Back to you.

pyledger/text_ledger.py
```diff
@@ -217,6 +225,18 @@ def write_ledger_file(df: pd.DataFrame, file: str) -> pd.DataFrame:
...
 
+        # Apply the smallest precision
+        def format_with_precision(series: pd.Series, precision: float) -> pd.Series:
+            decimal_places = -1 * math.floor(math.log10(precision))
+            return series.apply(lambda x: pd.NA if pd.isna(x) else f"{x:.{decimal_places}f}")
+
+        precision = df.apply(lambda row: self.precision(row["currency"], row["date"]), axis=1).min()
+        reporting_precision = df.apply(
+            lambda row: self.precision(self.reporting_currency, row["date"]), axis=1
+        ).min()
```
lasuk, line 236: There is only one reporting_currency (it is a global settings), no need to iterate over rows.

Directly use `self.precision("reporting_currency")` instead of `reporting_precision` further below on L238

lasuk (commented):
Thanks for the changes. One more comment.

pyledger/text_ledger.py
AlexTheWizardL: My silly mistake, thanks for pointing

AlexTheWizardL:
In commit [249871f](https://github.com/macxred/pyledger/pull/51/commits/249871fa7a7da5cc9df1c9f3c4f403fb77ccac13), test actions failed due to missing default asset loading when no assets were present. This may cause confusion if users are unable to write a valid ledger DataFrame without assets.

**Proposed Solutions:**

1. **Fallback to `DEFAULT_ASSETS` on Error:** For stability, load `DEFAULT_ASSETS` whenever asset reading fails. This keeps the system operational even if asset loading encounters issues, though it may complicate `clear` and `restore` logic.

2. **Generic Fallback in `precision()` Method:** Update `precision()` to return a default value if no assets are available, with a warning that a default was used. This approach keeps asset loading unchanged while providing a general solution.

3. **Isolated Solution in `write_ledger_file` for `TextLedger`:** Since this issue is specific to `TextLedger`, an isolated fix within `write_ledger_file` could set a default if no precision data is found, minimizing broader impact.

**Preferred Solution:** I favor the second option, as it provides a clean, generic fallback. Your feedback is appreciated.

AlexTheWizardL:
After spending some time working on the issue with float values, I realized that loading default assets when no file is present is not an ideal solution. Those default values would then be used within mutator and mirroring methods, which could lead to unexpected results.

Instead, I propose a more contained solution within TextLedger. Specifically, I suggest using a default value only within the write method when converting values to strings. This approach keeps the default handling encapsulated and avoids unintended usage in other methods.

pyledger/standalone_ledger.py
```diff
@@ -341,7 +341,11 @@ def precision(self, ticker: str, date: datetime.date = None) -> float:
 
         increment = self._assets_as_dict_of_df.get(ticker)
         if increment is None:
-            raise ValueError(f"No asset definition available for ticker '{ticker}'.")
+            if not DEFAULT_ASSETS["ticker"].unique():
+                raise ValueError("Default assets must contain unique ticker values.")
+            increment = DEFAULT_ASSETS.query(f"ticker == '{ticker}'")
+            if increment.empty:
+                raise ValueError(f"No asset definition available for ticker '{ticker}'.")
```
lasuk, line 348: Streamlined and more readable version:
```python
asset = self._assets_as_dict_of_df.get(ticker)
if asset is None:
    # Asset is not defined by the user, fall back to hard-coded defaults
    increment = DEFAULT_ASSETS.loc[DEFAULT_ASSETS["ticker"] == ticker, "increment"]
    if len(increment) < 1:
        raise ValueError(f"No asset definition available for ticker '{ticker}'.")
    if len(increment) > 1:
        raise ValueError(f"Multiple default definitions for asset '{ticker}'.")
    return increment.item()
else:
    mask = asset["date"].isna() | (asset["date"] <= pd.Timestamp(date))
    if not mask.any():
        raise ValueError(f"No asset definition available for '{ticker}' on or before {date}.")
    return asset.loc[mask[mask].index[-1], "increment"].item()
```


Intead of:
```python
increment = self._assets_as_dict_of_df.get(ticker)
if increment is None:
    if not DEFAULT_ASSETS["ticker"].unique():
        raise ValueError("Default assets must contain unique ticker values.")
    increment = DEFAULT_ASSETS.query(f"ticker == '{ticker}'")
    if increment.empty:
        raise ValueError(f"No asset definition available for ticker '{ticker}'.")
mask = increment["date"].isna() | (increment["date"] <= pd.Timestamp(date))
if not mask.any():
    raise ValueError(f"No asset definition available for '{ticker}' on or before {date}.")
return increment.loc[mask[mask].index[-1], "increment"]
```

(Efficiency is increased thanks to a single lookup in the default case; use `.item()` to return the value as float instead of pd.Series; clearly distinguish base and fall back cases)

lasuk (commented):
Thanks for the updates. Here's an additional comment.

pyledger/standalone_ledger.py
```diff
@@ -341,7 +341,11 @@ def precision(self, ticker: str, date: datetime.date = None) -> float:
 
         increment = self._assets_as_dict_of_df.get(ticker)
         if increment is None:
-            raise ValueError(f"No asset definition available for ticker '{ticker}'.")
+            if not DEFAULT_ASSETS["ticker"].unique():
```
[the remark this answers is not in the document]
AlexTheWizardL: I was sure that `.unique()` checks whether all values are unique, that was my mistake. `Unique` returns set of unique values and code wasn't completely right

PR #53, opened 2024-11-04
Refactor `StandaloneLedger`

AlexTheWizardL:
### Refactor `StandaloneLedger` to Remove Class Variables

This PR refactors `StandaloneLedger` by removing all class variables to ensure it functions purely as an abstract interface, without direct data storage or implementations for accessors or mutators.

**Changes in this PR:**
- Removed all class variables from `StandaloneLedger`.
- Introduced `_prices_as_dict_of_df`, aligning with the approach used for `_assets_as_dict_of_df` to optimize data handling.
- Revised the `complete_ledger()` method to eliminate local variable dependencies.

**Note:** The caching strategy remains challenging, particularly for cache invalidation across multiple elements (e.g., tax, prices). This PR includes an my input for discussion on how to best handle caching in this context. Feedback and ideas on improving cache management and invalidation would be highly appreciated.


pyledger/standalone_ledger.py
```diff
@@ -339,11 +141,10 @@ def complete_ledger(self) -> None:
             date=df.loc[index, "date"]
         )
 
-        # revaluations
+        # Revaluations
         revaluations = self.revaluations()
         reporting_currency = self.reporting_currency
         for row in revaluations.to_dict("records"):
-            self._serialized_ledger = self.serialize_ledger(df)
```
lasuk, line 346: With this change, `account_balance(..)` further below will return a different value.
I need to sleep over this before making a proposal.

pyledger/standalone_ledger.py
```diff
@@ -438,7 +235,7 @@ def _report_amount(
             raise ValueError("Vectors 'amount', 'currency', and 'date' must have the same length.")
         result = [
             self.round_to_precision(
-                a * self.price(t, date=d, currency=reporting_currency)[1],
+                a * 100,
```
lasuk, line 238: I do not understand this change.

lasuk (commented):
Thanks for converting StandaloneLedger to an abstract class.

We can delete even more unneeded code: I've added four more commits to drop or reshuffle methods.

I've left two review comments to discuss.

pyledger/standalone_ledger.py
lasuk: Fixed in 5dcbb2e:

- Compute account balance from ledger entries provided as argument to `_balance_from_serialized_ledger`.
- Isolate computation of revaluation entries

pyledger/standalone_ledger.py
AlexTheWizardL: This was made by a mistake. Should be rollbacked

AlexTheWizardL: UPD: rollbacked

pyledger/standalone_ledger.py
lasuk: This has been implemented and tested in the meantime.

lasuk:
Merging. Thanks for the excellent work!

PR #59, opened 2024-11-14
Standardize base tests

AlexTheWizardL:
This PR updates the test suites to align all entity base tests with a consistent, standard approach, with the **exception of the Ledger entity base test**.

During this work, I identified an opportunity for an improved architectural approach. Introducing a factory pattern for base tests, similar to what we implemented for class entities, would enhance flexibility and maintainability in the future.

### DIFF
--- a/pyledger/tests/base_test_revaluation.py
+++ b/pyledger/tests/base_test_revaluation.py
@@ -28,7 +28,7 @@ class BaseTestRevaluations(BaseTest):
             check_like=True, ignore_row_order=ignore_row_order
         )
 
-        # Modify with only id columns and one column to update
+        # Modify only a single column in a specific row
         revaluations.loc[0, "debit"] = 1005
         engine.revaluations.modify([{
             "account": revaluations.loc[0, "account"],
@@ -40,7 +40,7 @@ class BaseTestRevaluations(BaseTest):
             check_like=True, ignore_row_order=ignore_row_order
         )
 
-        # Modify with a DataFrame containing all columns from the schema
+        # Modify all columns from the schema in a specific row
         revaluations.loc[3, "description"] = "Modify with all columns test"
         engine.revaluations.modify([revaluations.loc[3]])
         assert_frame_equal(
@@ -56,7 +56,7 @@ class BaseTestRevaluations(BaseTest):
             check_like=True, ignore_row_order=ignore_row_order
         )
 
-        # Delete single row
+        # Delete a single row
         engine.revaluations.delete([{
             "account": revaluations['account'].iloc[0],
             "date": revaluations['date'].iloc[0],
--- a/pyledger/tests/base_test_revaluation.py
+++ b/pyledger/tests/base_test_revaluation.py
@@ -103,23 +103,28 @@ class BaseTestRevaluations(BaseTest):
         # Ensure the DataFrame passed as argument to revaluations.mirror() remains unchanged.
         assert_frame_equal(target, original_target, ignore_row_order=True)
         assert_frame_equal(
-            target, ledger.revaluations.list(),
-            ignore_row_order=True, check_like=True
+            target, ledger.revaluations.list(), ignore_row_order=True, check_like=True
         )
 
+        # Mirror with delete=False shouldn't change the data
         target = self.REVALUATIONS.query("credit not in [8050]")
+        ledger.revaluations.mirror(target, delete=False)
+        assert_frame_equal(
+            original_target, ledger.revaluations.list(), ignore_row_order=True, check_like=True
+        )
+
+        # Mirror with delete=True should change the data
         ledger.revaluations.mirror(target, delete=True)
         assert_frame_equal(
-            target, ledger.revaluations.list(),
-            ignore_row_order=True, check_like=True
+            target, ledger.revaluations.list(), ignore_row_order=True, check_like=True
         )
 
+        # Reshuffle target data randomly and modify one of the rows
         target = target.sample(frac=1).reset_index(drop=True)
         target.loc[target["account"] == "1000:2999", "description"] = "New Description"
         ledger.revaluations.mirror(target, delete=True)
         assert_frame_equal(
-            target, ledger.revaluations.list(),
-            ignore_row_order=True, check_like=True
+            target, ledger.revaluations.list(), ignore_row_order=True, check_like=True
         )
 
     def test_mirror_empty_revaluations(self, ledger):
--- a/pyledger/tests/base_test_revaluation.py
+++ b/pyledger/tests/base_test_revaluation.py
@@ -18,8 +18,8 @@ class BaseTestRevaluations(BaseTest):
     def test_revaluations_accessor_mutators(self, ledger, ignore_row_order=False):
         ledger.restore(settings=self.SETTINGS, accounts=self.ACCOUNTS)
 
-        # Add revaluations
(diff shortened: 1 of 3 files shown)


pyledger/tests/base_test_accounts.py
lasuk, line 1: The word 'ledger' has several meaning. We use it both to designate the accounting engine (an instance of the LedgerEngine base class), end for the general ledger or journal (the collection of accounting records). The first usage is uncommon and a bit misleading.

I propose to clarify our code and stop using `ledger` in the context of `accounting engine` and use `engine` as variable name instead of `ledger`. Please feel free to implement an alternative naming that avoids the double use of the word `ledger`.

Also, I suggest to rename the `LedgerEngine` base class to `AccountEngine`. But this is outside of the scope of the present PR.

pyledger/tests/base_test_accounts.py
```diff
@@ -0,0 +1,143 @@
...
+        pass
+
+    def test_account_accessor_mutators(self, ledger, ignore_row_order=False):
+        tax_accounts = self.ACCOUNTS.query("account in [4000, 5000]")
+        ledger.restore(tax_codes=self.TAX_CODES, accounts=tax_accounts, settings=self.SETTINGS)
+
+        # Add accounts
+        accounts = self.ACCOUNTS.query("account not in [4000, 5000]").sample(frac=1)
+        for account in accounts.to_dict('records'):
+            ledger.accounts.add([account])
```
lasuk, line 25: Please add some accounts one by one and others as a DataFrame holding multiple rows.

For example, you could loop over all rows except the last three and then add the last three rows (`tail(3).reset_index(FALSE)` at once.

(Applies to all base test classes)

pyledger/tests/base_test_accounts.py
```diff
@@ -0,0 +1,143 @@
...
+        for i in rows:
+            accounts.loc[i, "description"] = f"Modify Description: {i}"
+            ledger.accounts.modify({
+                "account": [accounts.loc[i, "account"]],
+                "description": [f"Modify Description: {i}"]
+            })
+            assert_frame_equal(
+                ledger.accounts.list(), accounts,
+                check_like=True, ignore_row_order=ignore_row_order
+            )
```
lasuk, line 52: Please test the following modify cases:
- Pass a single row with solely id columns and one column to `modify()`
- Pass a single row with all columns in the schema to the `modify()`
- Pass multiple rows to `modify()`

Pick one element at the beginning, some in the middle and one at the end to ensure all positions are covered.

Since these tests for the modify method are different, they are probably best written without a loop.

(Applies to all base test classes)

pyledger/tests/base_test_accounts.py
```diff
@@ -0,0 +1,143 @@
...
+
+        # Delete accounts
+        ledger.accounts.delete([{"account": accounts['account'].iloc[rows[0]]}])
+        accounts = accounts.drop(rows[0]).reset_index(drop=True)
+        assert_frame_equal(
+            ledger.accounts.list(), accounts, check_like=True, ignore_row_order=ignore_row_order
+        )
```
lasuk, line 59: Please test both deleting a single and deleting multiple entries. In the case of multiple entries, pleas pick to rows that are not adjacent.

(Applies to all base test classes)

pyledger/tests/base_test_accounts.py
```diff
@@ -0,0 +1,143 @@
...
+        assert_frame_equal(target, ledger.accounts.list(), ignore_row_order=True, check_like=True)
+
+        # Mirror with delete=False shouldn't change the data
+        target = self.ACCOUNTS.query("account not in [1000, 1005]")
+        ledger.accounts.mirror(target, delete=False)
+        assert_frame_equal(
+            original_target, ledger.accounts.list(), ignore_row_order=True, check_like=True
+        )
+
+        # Mirror with delete=False shouldn't change the data
```
lasuk, line 114: I believe this comment is incorrect.

lasuk (commented):
Thanks for streamlining and standardising the tests. 
I left a few review comments.

pyledger/tests/base_test_price_history.py
```diff
@@ -12,108 +12,142 @@ class BaseTestPriceHistory(BaseTest):
...
+        )
+
+        # Add prices as a DataFrame with a multiple rows
+        tail_prices = self.PRICES.tail(3).sample(frac=1)
+        engine.price_history.add(tail_prices)
+        price_history = pd.concat([price_history, tail_prices], ignore_index=True)
+        assert_frame_equal(
+            engine.price_history.list(), price_history,
+            check_like=True, ignore_row_order=ignore_row_order
+        )
```
lasuk, line 37: Shorter version of the same test:
```python
# Add
price_history = self.PRICES.sample(frac=1).reset_index(drop=True)
for price in price_history.head(-3).to_dict('records'):
    engine.price_history.add([price])
engine.price_history.add(price_history.tail(3))
assert_frame_equal(
    engine.price_history.list(), price_history,
    check_like=True, ignore_row_order=ignore_row_order
)
```
Instead of:
```
# Add prices one by one
price_history = self.PRICES.head(-3).sample(frac=1).reset_index(drop=True)
for price in price_history.to_dict('records'):
    engine.price_history.add([price])
assert_frame_equal(
    engine.price_history.list(), price_history,
    check_like=True, ignore_row_order=ignore_row_order
)

# Add prices as a DataFrame with a multiple rows
tail_prices = self.PRICES.tail(3).sample(frac=1)
engine.price_history.add(tail_prices)
price_history = pd.concat([price_history, tail_prices], ignore_index=True)
assert_frame_equal(
    engine.price_history.list(), price_history,
    check_like=True, ignore_row_order=ignore_row_order
)
```

(Applies to all entities)

pyledger/tests/base_test_price_history.py
```diff
@@ -12,108 +12,142 @@ class BaseTestPriceHistory(BaseTest):
...
     @abstractmethod
     @pytest.fixture
-    def ledger(self):
+    def engine(self):
         pass
 
-    def test_price_accessor_mutators(self, ledger, ignore_row_order=False):
-        ledger.restore(settings=self.SETTINGS)
-        price_history = self.PRICES.sample(frac=1).reset_index(drop=True)
+    def test_price_accessor_mutators(self, engine, ignore_row_order=False):
```
lasuk, line 18: Why is `ignore_row_order=False` by default? The test should be stronger by default and enforce `ignore_row_order=True`.

(applies to all entities)

lasuk (commented):
As discussed on our call, here's an incomplete review.

Thanks for narrowing the scope in a first step to a model case.

pyledger/tests/base_test_price_history.py
AlexTheWizardL: I believe it is correct now. Here is the part of the docstring of the assert_frame_equal() method:
```python
ignore_row_order (bool, optional): If True, the comparison will disregard the sequence
                                           in which rows appear.
```

Setting it to False makes the test stronger

AlexTheWizardL:
@lasuk I did narrow the scope of the PR to apply a standard approach only for `revaluations` entity

lasuk:
> @lasuk I did narrow the scope of the PR to apply a standard approach only for `revaluations` entity

Thanks! Much appreciated.

pyledger/tests/base_test_price_history.py
lasuk: You're right. Thanks. My mistake.

PR #61, opened 2024-11-15
Standardize base test except ledger

AlexTheWizardL:
This PR updates all base test suites, **except Ledger**, to align with the standardized format established in the previous PR for the revaluations base test.


pyledger/tests/base_test_accounts.py
```diff
@@ -0,0 +1,145 @@
...
+                "account": 77777, "currency": "CHF", "description": "test account"
+            }])
+
+    def test_delete_account_allow_missing(
+        self, engine, error_class=ValueError, error_message="Some ids are not present in the data."
+    ):
+        with pytest.raises(error_class, match=error_message):
+            engine.accounts.delete([{
+                "account": 77777, "currency": "CHF", "description": "test account"
+            }], allow_missing=False)
```
lasuk, line 95: This would be sufficient:
```
 engine.accounts.delete([{"account": 77777}], allow_missing=False)
```
Instead of:
```
 engine.accounts.delete([{
    "account": 77777, "currency": "CHF", "description": "test account"
}], allow_missing=False)
```

(also applies to the next line)

pyledger/tests/base_test_accounts.py
```diff
@@ -0,0 +1,145 @@
...
+
+    @abstractmethod
+    @pytest.fixture
+    def engine(self):
+        pass
+
+    def test_account_accessor_mutators(self, engine, ignore_row_order=False):
+        ACCOUNTS = self.ACCOUNTS.query("account not in [4000, 5000]")
+        tax_accounts = self.ACCOUNTS.query("account in [4000, 5000]")
+        engine.restore(tax_codes=self.TAX_CODES, accounts=tax_accounts, settings=self.SETTINGS)
```
lasuk, line 21: Why do we need the distinction into `ACCOUNTS` and `tax_accounts`?

pyledger/tests/base_test_accounts.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # Modify all columns from the schema in a specific row
+        accounts.loc[3, "description"] = "Modify with all columns test"
+        engine.accounts.modify([accounts.loc[3]])
+        assert_frame_equal(
+            engine.accounts.list(), accounts,
+            check_like=True, ignore_row_order=ignore_row_order
+        )
+
+        # Modify with a multiple rows
```
lasuk, line 51: Typo: "Modify multiple rows", rather than "Modify with a multiple rows"

pyledger/tests/base_test_tax_codes.py
```diff
@@ -0,0 +1,136 @@
...
+            }])
+
+    def test_delete_tax_code_allow_missing(
+        self, engine, error_class=ValueError, error_message="Some ids are not present in the data."
+    ):
+        with pytest.raises(error_class, match=error_message):
+            engine.tax_codes.delete([{
+                "id": "TestCode", "description": "tax 20%",
+                "account": 9990, "rate": 0.02, "is_inclusive": True
+            }], allow_missing=False)
```
lasuk, line 100: Can be shortened to:
```
engine.tax_codes.delete([{"id": "TestCode"}], allow_missing=False)
```

(applies also to next line)

pyledger/tests/base_test_accounts.py
AlexTheWizardL: The reason for including accounts in the `restore()` method is that they are required for Tax Codes during the restoring process. We have agreed that each test suite should define complete data for the system to ensure consistency.

**Possible approaches:**

1. **Keep it as it is now:** Continue including accounts with Tax Codes in the `restore()` method.
2. **Take accounts without Tax Codes:** Adjust the approach to exclude Tax Codes when restoring accounts.

lasuk: Thanks. Agreed to leave as is for now.

We can always review later when working on CashCtrlLedger if needed or appropriate.

PR #62, opened 2024-11-15
Standardize ledger base test

AlexTheWizardL:
This PR aligns and standardizes the ledger base test with the rest of the base tests, ensuring consistency across the test framework.

Key changes:
- **Dropped redundant tests:** Removed some tests (e.g., accessor/mutators without tax codes) as they appeared unnecessary. Please let me know if you feel we are missing essential test cases.
- **Streamlined transaction type logic:** Incorporated logic for converting transaction types (single/collective) directly within the accessor/mutators test. Suggestions for improving this implementation are welcome!

Feedback is appreciated, especially if there are concerns about missing test coverage or better ways to handle transaction type logic.

PR #63, opened 2024-11-16
Coherent naming in base tests

lasuk:
This PR follows up on #62 and align names in ledger base test with other base tests.

Changes:
1. Use `engine` rather than `pristine_engine` to name the fixture that instantiates the accounting engine, as in other base tests.
2. Rename fixture that represents a pre-populated accounting engine for testing to `restore_engine`.
3. Clear any existing ledger entries in `restore_engine`. This will be helpful when testing connections to external systems such as CashCtrl, that will not start from a pristine state.

### DIFF
--- a/pyledger/tests/base_test_ledger.py
+++ b/pyledger/tests/base_test_ledger.py
@@ -11,15 +11,19 @@ class BaseTestLedger(BaseTest):
 
     @abstractmethod
     @pytest.fixture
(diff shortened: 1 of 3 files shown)


AlexTheWizardL:
Thanks for the changes

AlexTheWizardL (approved):

PR #64, opened 2024-11-16
`restore_engine` fixture for accounts and tax codes 

AlexTheWizardL:
This PR proposes the introduction of a `restore_engine` method in base test classes (e.g., `BaseTestTaxCodes` and `BaseTestAccounts`) to ensure test coherence by prepopulating the engine with relevant data. This change aims to simplify test setup and enable generalization across similar use cases.

I changed the test flow for `tax_codes` accessor/mutators since we have only 5 taxes and after restoring 2 needed for accounts we only have 3 left, but it still covers the same cases

### DIFF
--- a/pyledger/tests/base_test_tax_codes.py
+++ b/pyledger/tests/base_test_tax_codes.py
@@ -17,20 +17,20 @@ class BaseTestTaxCodes(BaseTest):
     @pytest.fixture()
     def restored_engine(self, engine):
         """Accounting engine populated with accounts, tax codes, and settings"""
-        tax_codes = self.TAX_CODES.query("id in ['OUT_STD', 'IN_STD']")
-        engine.restore(accounts=self.ACCOUNTS, tax_codes=tax_codes, settings=self.SETTINGS)
+        tax_accounts = pd.concat([
+            self.TAX_CODES["account"], self.TAX_CODES["contra"]
+        ]).dropna().unique()
+        tax_accounts = self.ACCOUNTS.query("`account` in @tax_accounts")
+        engine.restore(accounts=tax_accounts, tax_codes=[], settings=self.SETTINGS)
         return engine
 
     def test_tax_codes_accessor_mutators(self, restored_engine, ignore_row_order=False):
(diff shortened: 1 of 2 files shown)


pyledger/tests/base_test_tax_codes.py
```diff
@@ -14,14 +14,23 @@ class BaseTestTaxCodes(BaseTest):
...
-    def test_tax_codes_accessor_mutators(self, engine, ignore_row_order=False):
-        engine.restore(accounts=self.ACCOUNTS, settings=self.SETTINGS)
+    @pytest.fixture()
+    def restored_engine(self, engine):
+        """Accounting engine populated with accounts, tax codes, and settings"""
+        tax_codes = self.TAX_CODES.query("id in ['OUT_STD', 'IN_STD']")
+        engine.restore(accounts=self.ACCOUNTS, tax_codes=tax_codes, settings=self.SETTINGS)
+        return engine
+
+    def test_tax_codes_accessor_mutators(self, restored_engine, ignore_row_order=False):
```
lasuk, line 24: Will be simplified if `restore_engine` clears all tax codes.

pyledger/tests/base_test_tax_codes.py
```diff
@@ -77,29 +87,34 @@ def test_create_existing__tax_code_raise_error(
...
+        error_message="Some ids are not present in the data."
     ):
         with pytest.raises(error_class, match=error_message):
-            engine.tax_codes.delete([{"id": "TestCode"}], allow_missing=False)
-        engine.tax_codes.delete([{"id": "TestCode"}], allow_missing=True)
+            restored_engine.tax_codes.delete([{"id": "TestCode"}], allow_missing=False)
+        restored_engine.tax_codes.delete([{"id": "TestCode"}], allow_missing=True)
 
-    def test_mirror_tax_codes(self, engine):
+    def test_mirror_tax_codes(self, restored_engine):
```
lasuk, line 112: Will be simplified if `restore_engine` clears all tax codes.

PR #67, opened 2024-11-21
 Add ExternalTabularEntity class

AlexTheWizardL:
This PR introduces the `ExternalTabularEntity` class, designed to facilitate seamless integration with external systems.

### DIFF
--- a/pyledger/__init__.py
+++ b/pyledger/__init__.py
@@ -37,3 +37,4 @@ from .tests import (
     BaseTestPriceHistory,
     BaseTestRevaluations,
 )
+from .storage_entity import AccountingEntity


pyledger/storage_entity.py
```diff
@@ -657,3 +657,19 @@ def delete(self, id: pd.DataFrame, allow_missing: bool = False):
         paths_to_update = self._csv_path(incoming["id"]).unique()
         for path in paths_to_update:
             self._store(new[self._csv_path(new["id"]) == path], self._path / path)
+
+
+class ExternalTabularEntity(AccountingEntity):
```
lasuk, line 662: Given the name change from `TabularEntity` to `AccountingEntity`, I would name the new class `ExternalAccountingEntity` instead of `ExternalTabularEntity`.

AlexTheWizardL:
`ExternalTabularEntity` shall be renamed to `CashCtrlAccountingEntity` and implemented in the `cashctrl_ledger` package.

lasuk:
I propose to define the new class directly in the cashctrl_ledger package and name it `CashCtrlAccountingEntity`.

Reasoning:
- Define the class where it is used (i.e. where child classes are defined) to make the usage more evident to readers.
- In `cashctrl_ledger`, we can type hint client as `CashCtrlAccountingEntity` instead of `Any`. The usage of Any in type hint somewhat undermines the objective pursued with type hints.

AlexTheWizardL:
Scope of this PR is now reduced to only import `AccountingEntity` class

PR #68, opened 2024-11-22
Comparsion NA values fix

AlexTheWizardL:
Comparisons involving `pd.NA` in Pandas DataFrames result in `pd.NA`, not boolean values, which causes methods like `.ne()` and `.any()` to miss differences involving missing values.

This update addresses the issue by separately comparing the missing value statuses using `.isna()` and combining the results with the value comparisons. This ensures that all differences—including those where values change to or from `pd.NA`—are accurately detected.

Do we need to test that case in all entities base tests?


lasuk:
Thanks for catching, explaining and fixing the problem!

lasuk:
> Do we need to test that case in all entities base tests?

Yes, I would appreciate if the same test was added to base tests of all entities.

Issue #71, opened 2024-11-27
Discard incoherent data


AlexTheWizardL:
Context: While refactoring StandalonLedger, we have inadvertently dropped the coherence check, which discards incoherent elements in accounting data before answering account_balance, account_history or similar request.
Goal: Re-introduce coherence check in StandalonLedger. Ensure it enforces coherence across all entities, including newly created ones.
Guiding Principles: 

1. StandaloneLedger child classes do not perform coherence checks when adding, modifying or deleting data. This is helpful when writing scripts to automatically modify the accounting data, as mutator functions will not rais an error for incoherent data.
2. Instead, the coherence check is performed before answering queries based on accounting data, such as the balance of a particular account. The coherence test will drop incoherent data with a warning message in the log (which is by default written out to the console).
3. The coherence test pursues the following goals:
    1. Ensure consistency of the retained data.
    2. Discard as little elements as possible. 
        (For example, if a ledger entry uses an unknown tax code, discard that tax code but not the entire ledger entry.)

 Examples:

* Discard ledger entries with a non-zero balance.
* Discard ledger entries with an unknown account (not defined in the account chart).
* If a ledger entry uses an unknown tax code, discard that tax code but not the entire ledger entry.
* etc.

AlexTheWizardL:
### List of rules (Need to be validated by Lukas):

**Updated Validation Rules for TAX_CODES:**

1. **Valid Account Reference (if rate other than 0)**: If the `rate` is not zero and `account` is specified, the account must exist in the `ACCOUNTS` entity.
2. **Non-negative Rate**: The `rate` must be a non-negative numeric value.

-----

**Updated Validation Rules for ACCOUNTS:**

1. **Valid Tax Code Reference**: If `tax_code` is specified, it must exist in the `TAX_CODES` entity.

-----

**Updated Validation Rules for LEDGER:**

1. **Existing Accounts**: If specified, `account` and `contra` must exist in the `ACCOUNTS` entity.
2. **Valid Tax Code Reference**: If `tax_code` is specified, it must exist in the `TAX_CODES` entity.
3. **Balanced Transactions**: For each `id`, the sum of `amount`s must balance to zero per currency.

-----

**Updated Validation Rules for ASSETS:**

1. **Positive Increment**: The `increment` must be a positive numeric value.

-----

**Updated Validation Rules for PRICES:**

1. **Existing Tickers**: The `ticker` must exist in the `ASSETS` entity.
[...]

PR #78, opened 2024-12-05
Cache serialized ledger

AlexTheWizardL:
This PR goal is to add caching for the `serialized_ledger()` method and design a mechanism for invalidating it.

### DIFF
--- a/pyledger/tests/test_serialized_ledger_cache.py
+++ b/pyledger/tests/test_serialized_ledger_cache.py
@@ -1,0 +1,38 @@
+"""Unit tests for the serialized ledger caching mechanism."""
+
+import pytest
+from consistent_df import assert_frame_equal
+from .base_test import BaseTest
+from pyledger import MemoryLedger
+
+
+def test_serialized_ledger_cache():
+    engine = MemoryLedger()
+    engine.restore(
+        settings=BaseTest.SETTINGS, tax_codes=BaseTest.TAX_CODES,
+        accounts=BaseTest.ACCOUNTS, ledger=BaseTest.LEDGER_ENTRIES, price_history=BaseTest.PRICES
+    )
+
+    # Serialized ledger should change when removing tax codes used in ledger entries.
+    serialized_ledger = engine.serialized_ledger()
+    engine._tax_codes._df = BaseTest.TAX_CODES.head(2)
+    assert_frame_equal(engine.serialized_ledger(), serialized_ledger, ignore_row_order=True)
+    engine.serialized_ledger.cache_clear()
+    with pytest.raises(AssertionError):
+        assert_frame_equal(engine.serialized_ledger(), serialized_ledger, ignore_row_order=True)
+
+    # Serialized ledger should change when removing accounts used in ledger entries.
+    serialized_ledger = engine.serialized_ledger()
+    engine.accounts._df = BaseTest.ACCOUNTS.head(5)
+    assert_frame_equal(engine.serialized_ledger(), serialized_ledger, ignore_row_order=True)
+    engine.serialized_ledger.cache_clear()
+    with pytest.raises(AssertionError):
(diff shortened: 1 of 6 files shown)


pyledger/tests/test_serialized_ledger_cache.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def test_serialized_ledger_cache():
+    engine = MemoryLedger()
+    engine.restore(
+        settings=BaseTest.SETTINGS, tax_codes=BaseTest.TAX_CODES,
+        accounts=BaseTest.ACCOUNTS, ledger=BaseTest.LEDGER_ENTRIES, price_history=BaseTest.PRICES
+    )
+
+    # Serialized ledger should change when removing tax codes used in ledger entries.
```
lasuk, line 16: The comment is unclear to me. What do we test, and what is the motivation for this code block?

pyledger/tests/test_serialized_ledger_cache.py
```diff
@@ -0,0 +1,38 @@
...
+    serialized_ledger = engine.serialized_ledger()
+    engine._tax_codes._df = BaseTest.TAX_CODES.head(2)
+    assert_frame_equal(engine.serialized_ledger(), serialized_ledger, ignore_row_order=True)
+    engine.serialized_ledger.cache_clear()
+    with pytest.raises(AssertionError):
+        assert_frame_equal(engine.serialized_ledger(), serialized_ledger, ignore_row_order=True)
```
lasuk, line 22: A bit a convoluted way to check that two data frames are not equal. 
Can you check for an alternative?

pyledger/tests/test_serialized_ledger_cache.py
```diff
@@ -0,0 +1,38 @@
...
+def test_serialized_ledger_cache():
+    engine = MemoryLedger()
+    engine.restore(
+        settings=BaseTest.SETTINGS, tax_codes=BaseTest.TAX_CODES,
+        accounts=BaseTest.ACCOUNTS, ledger=BaseTest.LEDGER_ENTRIES, price_history=BaseTest.PRICES
+    )
+
+    # Serialized ledger should change when removing tax codes used in ledger entries.
+    serialized_ledger = engine.serialized_ledger()
+    engine._tax_codes._df = BaseTest.TAX_CODES.head(2)
```
lasuk, line 18: Dropping some tax codes will lead to warning messages once #71 is implemented.
I suggest to modify an existing tax code instead and then check whether the original or modified value is present in `serialised_ledger()`.

lasuk: Also applies to accounts  further below.

pyledger/tests/test_serialized_ledger_cache.py
lasuk: Also applies to accounts and ledger entries further below.

pyledger/tests/test_serialized_ledger_cache.py
```diff
@@ -0,0 +1,38 @@
...
+    assert_frame_equal(engine.serialized_ledger(), serialized_ledger, ignore_row_order=True)
+    engine.serialized_ledger.cache_clear()
```
lasuk, line 20: I propose to actually use `engine.tax_code.modify()` to check whether cache invalidation is properly triggered. Ideally we would also check that `engine.tax_code.add()` and `delete()` trigger cache invalidation

lasuk: Also applies to accounts and ledger entries further below.

lasuk:
All review comments are centred around a single test unit.
@AlexTheWizardL: I've pushed commit d9d627a with a sketch of how these comments could be addressed.
Build on this commit if you like. Ignore it otherwise.

PR #79, opened 2024-12-10
Keep extra columns functionality

AlexTheWizardL:
This PR introduces a mechanism to preserve extra columns when adding or modifying entity data.

### DIFF
--- a/pyledger/storage_entity.py
+++ b/pyledger/storage_entity.py
@@ -65,8 +65,7 @@ class AccountingEntity(ABC):
         Raises:
             ValueError: If required columns are missing or data types are incorrect.
         """
-        keep_extra_columns = not drop_extra_columns
-        df = enforce_schema(data, self._schema, keep_extra_columns=keep_extra_columns)
+        df = enforce_schema(data, self._schema, keep_extra_columns=not drop_extra_columns)
 
         # Convert -0.0 to 0.0
         for col in df.columns:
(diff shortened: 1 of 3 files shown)


pyledger/storage_entity.py
```diff
@@ -585,7 +590,7 @@ def write_directory(self, df: pd.DataFrame | None = None):
         if df is None:
             df = self.list()
 
-        df = self.standardize(df)
+        df = self.standardize(df).copy()
```
lasuk, line 593: Why do we need to copy after standardizing?

pyledger/tests/test_text_ledger_accounts.py
```diff
@@ -1,12 +1,51 @@
...
     def engine(self, tmp_path):
         return TextLedger(tmp_path)
+
+    def test_extra_columns(self, engine):
+        extra_cols = ["new_column", "second_new_column"]
+        accounts = self.ACCOUNTS.head(3).copy()
+        engine.accounts.add(accounts)
+
+        # Add accounts with a new column
+        expected = self.ACCOUNTS.tail(len(self.ACCOUNTS) - 3).copy()
```
lasuk, line 22: I believe this can be shortened to:
```
expected = self.ACCOUNTS.tail(-3).copy()
```

Instead of:
```
expected = self.ACCOUNTS.tail(len(self.ACCOUNTS) - 3).copy()
```

pyledger/tests/test_text_ledger_accounts.py
```diff
@@ -1,12 +1,51 @@
...
+        expected[extra_cols[0]] = "test value"
+        engine.accounts.add(expected)
+        current = engine.accounts.list()
+        account_numbers = accounts["account"].to_list()  # noqa: F841
+        assert current.query("account in @account_numbers")[extra_cols[0]].isna().all(), (
```
lasuk, line 27: How about
```
assert current.query("account in @accounts['account']")[extra_cols[0]].isna().all()
```
Instead of:
```
account_numbers = accounts["account"].to_list()  # noqa: F841
assert current.query("account in @account_numbers")[extra_cols[0]].isna().all()
```

pyledger/storage_entity.py
lasuk: Thanks.

PR #91, opened 2024-12-16
Sanitize assets and prices

AlexTheWizardL:
This PR goal is to implement sanitize_ assets and price methods

### DIFF
--- a/pyledger/ledger_engine.py
+++ b/pyledger/ledger_engine.py
@@ -784,27 +784,42 @@ class LedgerEngine(ABC):
             pd.DataFrame: The sanitized DataFrame with valid price entries.
         """
         df = enforce_schema(df, PRICE_SCHEMA, keep_extra_columns=True)
-        id_columns = PRICE_SCHEMA.query("id == True")["column"].tolist()
 
-        def validate_asset_reference(column):
-            """Validates specified column values for asset references"""
-            def is_valid(row):
+        # The `precision` method validates the existence of a valid asset definition
+        # and relies on sanitized data, serving as a centralized validation point.
+        def invalid_asset_reference(ticker: pd.Series, date: pd.Series) -> pd.Series:
+            """Validate specified column values for asset references."""
+            def is_valid(ticker, date):
                 try:
-                    self.precision(ticker=row[column], date=row["date"])
+                    self.precision(ticker=ticker, date=date)
                     return True
-                except ValueError as e:
-                    self._logger.warning(
-                        f"Discard price entry ({row[id_columns].to_dict()}): Invalid {column}: {e}"
-                    )
+                except ValueError:
                     return False
+            return pd.Series([is_valid(ticker=t, date=d) for t, d in zip(ticker, date)])
 
-            return df.apply(is_valid, axis=1)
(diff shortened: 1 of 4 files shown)


pyledger/ledger_engine.py
```diff
@@ -769,6 +771,43 @@ def round_scalar(amount: float, ticker: str, date: datetime.date = None) -> floa
...
+            df (pd.DataFrame): The raw prices DataFrame.
+            keep_extra_columns (bool): If True, retains columns not defined in the schema.
+
+        Returns:
+            pd.DataFrame: The sanitized prices DataFrame.
+        """
+        df = enforce_schema(df, PRICE_SCHEMA, keep_extra_columns=keep_extra_columns)
+        id_columns = PRICE_SCHEMA.query("id == True")["column"].tolist()
+
+        def validate_asset_reference(column):
```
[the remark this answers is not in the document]
lasuk: I would then log a single warning message for invalid tickers and for invalid currencies.
Along the lines of:
```
"Discard 123 price entries with invalid tickers ["BLA", "BLI", "BLO", "BLU", "BLE", ...]."
"Discard 51 price entries with invalid currencies ["CAD", "TYC", "ABC"]."
```
(I typically abbreviate lists in warning messages after three or five entries and use three dots to indicate that the list was truncated).

pyledger/tests/test_sanitize.py
```diff
@@ -0,0 +1,86 @@
+"""Test suit for testing sanitize ledger operations"""
```
lasuk: Nice and concise test suite!
Thanks a lot!

pyledger/ledger_engine.py
```diff
@@ -769,6 +771,57 @@ def round_scalar(amount: float, ticker: str, date: datetime.date = None) -> floa
...
+                except ValueError:
+                    return False
+            return pd.Series([is_valid(ticker=t, date=d) for t, d in zip(ticker, date)])
+
+        ticker_mask = invalid_asset_reference(df["ticker"], df["date"])
+        currency_mask = invalid_asset_reference(df["currency"], df["date"])
+
+        if not ticker_mask.all():
+            invalid_tickers = df.loc[~ticker_mask, "ticker"].unique().tolist()
+            truncated_tickers = (invalid_tickers[:3] + ["..."]
```
lasuk, line 805: I propose to add a helper function to the utilities module that takes a list (or list like) and returns the `n` first elements as a comma separated string, followed by "..." if the list is longer than n elements.

Sketch:
```
def first_elements_as_str(x: List, n: int=5) -> str:
    """
    String preview of a list.
    
    Return a comma-separated string of the first n elements of x, appending
    "..." if x contains more than n elements. This can be useful for including
    a concise list preview in error messages or logs.
    """
    result = [str(i) for i in x[:n]]
    if len(x) > n:
        result.append("...")
    return ', '.join(result)
```

lasuk: Below message then becomes:
```
 f"Discard {sum(invalid_tickers)} price entries with invalid tickers: {first_elements_as_str(invalid_tickers).}."
 ```

PR #98, opened 2025-01-02
Define Sanitize ledger method

AlexTheWizardL:
This PR goal is to define a method that sanitizes ledger entries

### DIFF
--- a/pyledger/ledger_engine.py
+++ b/pyledger/ledger_engine.py
@@ -2,6 +2,7 @@
 
 from abc import ABC, abstractmethod
 import datetime
+from decimal import Decimal
 import logging
 import math
 import zipfile
@@ -814,23 +815,22 @@ class LedgerEngine(ABC):
             )
             df = df.query("id not in @invalid_ids")
 
-
         # Validate balancing amounts
         effective_amounts = df["report_amount"].copy()
         index = effective_amounts.isna()
-
-        # Perform this only if amount if not 0 because then will be -0
         effective_amounts.loc[index] = self.report_amount(
             amount=df.loc[index, "amount"],
             currency=df.loc[index, "currency"],
             date=df.loc[index, "date"]
         )
-        effective_amounts.loc[df["contra"].notna()] *= -1
-        effective_amounts.loc[df["account"].notna() & df["contra"].notna()] = 0
-        net_amounts = effective_amounts.groupby(df["id"]).transform("sum")
-        invalid_amounts_mask = net_amounts.ne(0) | net_amounts.ne(-0)
+        effective_amounts = effective_amounts.mask(df["contra"].notna() & df["account"].notna(), 0)
+        effective_amounts = effective_amounts.mask(
+            df["contra"].notna() & df["account"].isna(), -effective_amounts
+        )
+        effective_amounts = effective_amounts.apply(lambda x: Decimal(str(x)))
+        invalid_amounts_mask = ~effective_amounts.groupby(df["id"]).sum().eq(0)
         if invalid_amounts_mask.any():
-            invalid_ids = df.loc[invalid_amounts_mask, "id"].unique().tolist()
+            invalid_ids = invalid_amounts_mask[invalid_amounts_mask].index.unique().tolist()
             self._logger.warning(
                 f"Discarding {len(invalid_ids)} ledger entries where "
                 f"amounts do not balance to zero: {first_elements_as_str(invalid_ids)}"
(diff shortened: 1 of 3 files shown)


pyledger/ledger_engine.py
```diff
@@ -691,93 +691,190 @@ def serialized_ledger(self) -> pd.DataFrame:
...
+        invalid_tax_code_mask = (
+            ~df["tax_code"].isna() & ~df["tax_code"].isin(set(tax_codes["id"]))
+        )
+        if invalid_tax_code_mask.any():
+            invalid_ids = df.loc[invalid_tax_code_mask, "id"].unique().tolist()
+            self._logger.warning(
+                f"Discarding {len(invalid_ids)} ledger entries with invalid tax codes: "
+                f"{first_elements_as_str(invalid_ids)}"
+            )
+            df = df.query("id not in @invalid_ids")
```
lasuk, line 747: We should simply drop invalid entries in the ledger["tax_code"] column but leave the ledger entries otherwise unchanged. (This is the least invasive way to make data coherent in the presence of undefined tax codes).

pyledger/tests/base_test_dump_restore_clear.py
```diff
@@ -14,11 +14,15 @@ def engine(self):
         pass
 
     def test_restore(self, engine):
+        LEDGER_ENTRIES = self.LEDGER_ENTRIES.copy()
+        default_profit_center = self.PROFIT_CENTERS.iloc[0]["profit_center"]
+        LEDGER_ENTRIES["profit_center"] = \
+            LEDGER_ENTRIES["profit_center"].fillna(default_profit_center)
```
lasuk, line 20: I assume this code is temporary and will be reverted once the test data in `self.LEDGER_ENTRIES` is extended with profit centres?

If you agree, we should insert a TODO to revert this code later.

pyledger/tests/base_test_dump_restore_clear.py
```diff
@@ -41,16 +45,20 @@ def test_restore(self, engine):
...
+        target = engine.txn_to_str(LEDGER_ENTRIES).values()
         actual = engine.txn_to_str(engine.ledger.list()).values()
         assert sorted(target) == sorted(actual), "Targeted and actual ledger differ"
 
     def test_dump_and_restore_zip(self, engine, tmp_path):
         # Populate with test data
+        LEDGER_ENTRIES = self.LEDGER_ENTRIES.copy()
+        default_profit_center = self.PROFIT_CENTERS.iloc[0]["profit_center"]
+        LEDGER_ENTRIES["profit_center"] = \
+            LEDGER_ENTRIES["profit_center"].fillna(default_profit_center)
```
lasuk, line 57: Same as above.

pyledger/tests/test_sanitize.py
```diff
@@ -255,3 +255,130 @@ def test_sanitized_accounts_tax_codes(engine, capture_logs):
...
+    )
+    ledger = pd.read_csv(StringIO(LEDGER_CSV), skipinitialspace=True)
+    engine.restore(accounts=accounts_df, tax_codes=tax_df, price_history=prices, assets=assets)
+
+    # Sanitize ledger entries without profit centers
+    expected_ledger_df = engine.ledger.standardize(expected_ledger)
+    sanitized = engine.sanitize_ledger(engine.ledger.standardize(ledger))
+    assert_frame_equal(expected_ledger_df, sanitized)
+    log_messages = capture_logs.getvalue().strip().split("\n")
+    assert len(log_messages) > 0
```
lasuk, line 374: I suggest to test for the expected number of log messages (i.e. if all tests expected to fail actually generate a log message).

pyledger/tests/test_sanitize.py
```diff
@@ -255,3 +255,130 @@ def test_sanitized_accounts_tax_codes(engine, capture_logs):
...
+    log_messages = []
+
+    # Sanitize ledger entries with profit centers
+    PROFIT_CENTERS = pd.read_csv(StringIO(PROFIT_CENTERS_CSV), skipinitialspace=True)
+    engine.restore(profit_centers=PROFIT_CENTERS)
+    expected_ledger_df = engine.ledger.standardize(expected_ledger_with_profit_centers)
+    sanitized = engine.sanitize_ledger(engine.ledger.standardize(ledger))
+    assert_frame_equal(expected_ledger_df, sanitized)
+    log_messages = capture_logs.getvalue().strip().split("\n")
+    assert len(log_messages) > 0
```
lasuk, line 384: Same as above.

lasuk (commented):
Thanks for your endurance with this lengthy (but important) method and the associated tests!

lasuk:
Btw: When applying multiple filters (masks) to a data frame, it is more efficient to first combine all filters and then to filter the data frame only once. Filtering or subsetting a data frame is a potentially costly operation, as potentially large memory blocks need to be reassigned.

There is nothing we need to do, though. Above argument is not important in our particular use case, as we do not subset the data frame in the most common use case, when all data is coherent. I just wanted to ensure everyone is aware of the general case.

PR #99, opened 2025-01-03
Profit centers for account balance and history

AlexTheWizardL:
This PR goal is to implement profit center filtering logic for account balance and account history.

### DIFF
--- a/pyledger/tests/base_test_accounts.py
+++ b/pyledger/tests/base_test_accounts.py
@@ -157,13 +157,15 @@ class BaseTestAccounts(BaseTest):
         restored_engine.restore(
             accounts=self.ACCOUNTS, settings=self.SETTINGS, tax_codes=self.TAX_CODES,
             ledger=self.LEDGER_ENTRIES, assets=self.ASSETS, price_history=self.PRICES,
-            revaluations=self.REVALUATIONS
+            revaluations=self.REVALUATIONS, profit_centers=self.PROFIT_CENTERS
         )
         for _, row in self.EXPECTED_BALANCE.iterrows():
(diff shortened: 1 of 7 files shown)


pyledger/tests/base_test_accounts.py
```diff
@@ -157,13 +157,32 @@ def test_account_balance(self, restored_engine):
         restored_engine.restore(
             accounts=self.ACCOUNTS, settings=self.SETTINGS, tax_codes=self.TAX_CODES,
             ledger=self.LEDGER_ENTRIES, assets=self.ASSETS, price_history=self.PRICES,
-            revaluations=self.REVALUATIONS
+            revaluations=self.REVALUATIONS, profit_centers=self.PROFIT_CENTERS
         )
-        for _, row in self.EXPECTED_BALANCE.iterrows():
+        # Test account balance with specified profit centers
+        EXPECTED_BALANCE_WITH_PROFIT_CENTERS = self.EXPECTED_BALANCE.query("profit_center.notna()")
```
lasuk, line 163: Why do we exclude expected balances with missing (NA) profit centes?

lasuk (commented):
Thanks for implementing `account_balance` and `account_history` with profit centres.

I only have a single comment.

pyledger/tests/base_test_accounts.py
AlexTheWizardL: In this way, I declared how we can separate those concepts (profit centers balances / account balances).
But now I realize we can make a test more robust just by using `self.EXPECTED_BALANCE` there.
Thanks for this catch!

Issue #101, opened 2025-01-03
Test `account_history()` method


AlexTheWizardL:
Need to test `account_history()` method similarly how now tested `account_balance()` using defined `BaseTest` data

PR #128, opened 2025-02-06
Add `drop` Parameter to `account_history`

AlexTheWizardL:
This PR goal is to add `drop` Parameter to `account_history()` method that removes columns that do not provide meaningful information in the given context.

### DIFF
--- a/pyledger/ledger_engine.py
+++ b/pyledger/ledger_engine.py
@@ -603,9 +603,8 @@ class LedgerEngine(ABC):
             ):
                 df = df.drop(columns=["report_amount", "report_balance"])
             mandatory = ACCOUNT_SCHEMA.query("mandatory == True")["column"].tolist()
-            remove = df.columns.difference(mandatory)
-            if df.empty or df[remove].isna().all().all():
-                df = df.drop(columns=remove)
+            remove = [col for col in df.columns.difference(mandatory) if df[col].isna().all()]
+            df = df.drop(columns=remove)
 
         return df
 
--- a/pyledger/ledger_engine.py
+++ b/pyledger/ledger_engine.py
@@ -548,7 +548,7 @@ class LedgerEngine(ABC):
 
     def account_history(
         self, account: int | str | dict, period: datetime.date = None,
-        profit_centers: list[str] | str = None, drop: bool = False
+        profit_centers: list[str] | str = None, drop: bool = True
     ) -> pd.DataFrame:
         """Transaction and balance history of an account or a list of accounts.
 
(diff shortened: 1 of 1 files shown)


AlexTheWizardL:
This PR does not include tests for the added functionality. Tests will be added while implementing macxred/pyledger#101.

pyledger/ledger_engine.py
```diff
@@ -599,29 +602,45 @@ def account_history(
...
+            else:
+                accounts = self.accounts.list().query("account in @accounts")
+            if (
+                accounts["currency"].nunique() == 1
+                and accounts["currency"].iloc[0] == self.reporting_currency
+            ):
+                df = df.drop(columns=["report_amount", "report_balance"])
+            optional = ACCOUNT_SCHEMA.query("mandatory == False")["column"].tolist()
+            optional_existing = df.columns.intersection(optional)
+            if df.empty or df[optional_existing].isna().all().all():
```
lasuk, line 640: Suggestion for alternative logic: Drop all columns that contain only NA values and that are not mandatory.
This will also drop additional columns (outside of the schema) that are all NA.

pyledger/ledger_engine.py
```diff
@@ -579,18 +579,21 @@ def account_history(
             profit_centers: (list[str], str): Filter for journal entries. If not None, the result is
                                               calculated only from journal entries assigned to one
                                               of the profit centers in the filter.
+            drop (bool, optional): If True, drops columns that do not provide meaningful
+                                   information in the given context. Defaults to False.
```
lasuk, line 583: Add a sentence or two with a concrete description of columns to drop. For example:

```
drop (bool, optional): If True, drops columns that do not provide meaningful
    information in the given context: Drops columns containing only NA values,
    the "account" column if the history  is shown for a single account, 
    "report_amount" and "report_balance" if the history is shown solely for 
    accounts in reporting currency.
```

pyledger/ledger_engine.py
```diff
@@ -607,29 +617,40 @@ def account_history(
...
                 accounts, start=start, end=end, profit_centers=profit_centers
             )
         else:
             raise ValueError(
                 f"Account(s) '{account}' of type {type(account).__name__} not identifiable."
             )
 
-        return out
+        if drop:
+            if not df.empty and df['account'].nunique() == 1:
```
lasuk, line 643: To comply with the doc-string, you should check whether we allow only a single account to be displayed, not whether the data only contains a single account.

lasuk:
Added two commits:
- 226a480 clarifies the doc string
- fc7b89e fixes the logic to drop non-mandatory columns with all NA values: 
  We now drop each non-mandatory column that contains only NA values, rather than all non-mandatory columns, if they *all* only contain NA values.

PR #133, opened 2025-02-12
Fix error writing empty ledger file

AlexTheWizardL:
This PR goal is to fix error writing empty ledger file.


lasuk:
Merged. Thanks a lot for the changes!

PR #135, opened 2025-03-03
Account and contra are NA values in revaluation entries

AlexTheWizardL:
This PR goal is to fix revaluation entries that have both account and contra fields are NA.


lasuk:
Thanks! Merging ...

PR #137, opened 2025-03-06
Test `account_history()` method

AlexTheWizardL:
This PR goal is to test `account_history()` method.

### DIFF
--- a/pyledger/constants.py
+++ b/pyledger/constants.py
@@ -110,6 +110,23 @@ AGGREGATED_BALANCE_SCHEMA = pd.read_csv(
     StringIO(AGGREGATED_BALANCE_SCHEMA_CSV), skipinitialspace=True
 )
 
+ACCOUNT_HISTORY_SCHEMA_CSV = """
+    column,              dtype,                mandatory,       id
+    date,                datetime64[ns],       False,        False
+    account,             Int64,                False,        False
+    contra,              Int64,                False,        False
+    currency,            string[python],       False,        False
+    amount,              Float64,              False,        False
+    balance,             Float64,              False,        False
+    report_amount,       Float64,              False,        False
+    report_balance,      Float64,              False,        False
+    tax_code,            string[python],       False,        False
+    profit_center,       string[python],       False,        False
+    description,         string[python],       False,        False
+    document,            string[python],       False,        False
+"""
+ACCOUNT_HISTORY_SCHEMA = pd.read_csv(StringIO(ACCOUNT_HISTORY_SCHEMA_CSV), skipinitialspace=True)
+
 DEFAULT_PRECISION = 0.01
 
 DEFAULT_ASSETS = enforce_schema(
--- a/pyledger/constants.py
+++ b/pyledger/constants.py
@@ -112,6 +112,7 @@ AGGREGATED_BALANCE_SCHEMA = pd.read_csv(
 
 ACCOUNT_HISTORY_SCHEMA_CSV = """
     column,              dtype,                mandatory,       id
+    id,                  string[python],       False,         True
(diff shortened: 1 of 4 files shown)


AlexTheWizardL:
This data model, using a list of objects, was chosen because the CSV would be too wide if we applied the same model as **EXPECTED_BALANCES** (which uses a hardcoded string CSV DataFrame). Additionally, not all columns will be present, as the `drop=True` parameter also adds complexity when defining the CSV string DataFrame.  

I aimed to find the most optimal cases, balancing a variety of parameters and the history output length. If you have any suggestions for improvements, feel free to share!

pyledger/constants.py
```diff
@@ -110,6 +110,23 @@
...
+    amount,              Float64,              False,        False
+    balance,             Float64,              False,        False
+    report_amount,       Float64,              False,        False
+    report_balance,      Float64,              False,        False
+    tax_code,            string[python],       False,        False
+    profit_center,       string[python],       False,        False
+    description,         string[python],       False,        False
+    document,            string[python],       False,        False
+"""
+ACCOUNT_HISTORY_SCHEMA = pd.read_csv(StringIO(ACCOUNT_HISTORY_SCHEMA_CSV), skipinitialspace=True)
```
lasuk, line 128: To my understanding, ACCOUNT_HISTORY_SCHEMA should also hold an 'id' column.

After that, the ACCOUNT_HISTORY_SCHEMA is an extension of the JOURNAL_SCHEMA with additional `balance` and `report_balance` column. What do you think of defining it via:
```
# Define additional entries not in JOURNAL_SCHEMA
EXTRA_FIELDS_CSV = """
    column,              dtype,                mandatory,       id
    balance,             Float64,              False,        False
    report_balance,      Float64,              False,        False
"""
EXTRA_FIELDS = pd.read_csv(StringIO(EXTRA_FIELDS_CSV), skipinitialspace=True)
ACCOUNT_HISTORY_SCHEMA = pd.concat([JOURNAL_SCHEMA, EXTRA_FIELDS], ignore_index=True)

# Reorder extra columns
order = JOURNAL_SCHEMA["column"].to_list()
order.insert(order.index("amount") + 1, "balance")
order.insert(order.index("report_amount") + 1, "report_balance")
ACCOUNT_HISTORY_SCHEMA = (
    ACCOUNT_HISTORY_SCHEMA
    .set_index("column", drop=True)
    .loc[order]
    .reset_index()
)
```

pyledger/ledger_engine.py
lasuk, line 1: In `account_history`, please change the default for `drop` to `False`.

By defaulting to drop=False, we ensure consistent return types unless a user explicitly opts for the convenience option `drop` with a more ‘concise’ but type-changing behavior. This makes the interface safer and more predictable.

Propsal:
```
def account_history(
    self,
    account: int | str | dict,
    period: datetime.date = None,
    profit_centers: list[str] | str = None,
    drop: bool = False
) -> pd.DataFrame:
```
Instead of:
```
def account_history(
    self,
    account: int | str | dict,
    period: datetime.date = None,
    profit_centers: list[str] | str = None,
    drop: bool = True
) -> pd.DataFrame:
```

pyledger/tests/base_test_accounts.py
lasuk, line 1: Please change `account_history ` both with `drop=True` and `drop=False`.

You can easily achieve this be adding a `drop` argument to `format_expected_df(..)`. No need to define additional expected outcomes.

lasuk:
Thanks for adding a test for `account_history`. This is really valuable.

I've added a few review comments. Back to you, @AlexTheWizardL.

PR #143, opened 2025-03-24
Split `sanitize_journal` method into more readable chunks

AlexTheWizardL:
This PR goal is to split sanitize_journal method into more readable chunks.

### DIFF
--- a/pyledger/helpers.py
+++ b/pyledger/helpers.py
@@ -159,7 +159,7 @@ def first_elements_as_str(x: List[Any], n: int = 5) -> str:
     """
     if not x:
         return ""
-    result = [str(i) for i in x[:n]]
+    result = [str(i) for i in list(x)[:n]]
     if len(x) > n:
         result.append("...")
     return ", ".join(result)
--- a/pyledger/ledger_engine.py
+++ b/pyledger/ledger_engine.py
@@ -821,7 +821,9 @@ class LedgerEngine(ABC):
         """
         return self.serialize_ledger(self.journal.list())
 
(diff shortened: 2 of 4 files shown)


AlexTheWizardL:
**Additional finding:**

- Applying `apply(func, axis=1)` to a **non-empty** DataFrame returns a **Series**.  
- Applying it to an **empty** DataFrame returns an **empty DataFrame**, not a Series.  
- Adding `result_type='reduce'` ensures a **Series is always returned**, even if the DataFrame is empty.

<details>
  <summary>Click to see the code snippet</summary>

  ```python
  from io import StringIO
  import pandas as pd

  JOURNAL_CSV = """
      id,        date, account, contra, currency,      amount,
      1,  2024-01-01,    1000,       ,      USD,   800000.00,
      2,  2024-01-01,        ,   1000,      CHF,   800000.00,
      3,  2024-01-01,    1000,   2000,      USD,   800000.00,
  """
  JOURNAL = pd.read_csv(StringIO(JOURNAL_CSV), skipinitialspace=True)

  def invalid_price(row):
      if row["currency"] == "USD":
          return True
      else:
          return False

  invalid_mask = JOURNAL.apply(invalid_price, axis=1)
  print(invalid_mask)
  # 0     True
  # 1    False
  # 2     True
  # dtype: bool

  invalid_mask_2 = pd.DataFrame({}).apply(invalid_price, axis=1)
  print(invalid_mask_2)
  # Empty DataFrame
  # Columns: []
  # Index: []

  invalid_mask_3 = pd.DataFrame({}).apply(invalid_price, axis=1, result_type='reduce')
  print(invalid_mask_3)
  # Series([], dtype: float64)

PR #151, opened 2025-04-15
Clarify name and docstring for account balance methods

AlexTheWizardL:
This PR goal is to clarify name and docstring for account balance methods.


AlexTheWizardL:
Renamed `account_balances` to `individual_account_balances` to clarify that it returns one row per account, with each balance calculated independently and without aggregation.

Also updated the docstring for `account_balance` to emphasize that it returns a summarized balance for a single account, a range, or an account expression.

These changes improve clarity, align method names with actual behavior, and help distinguish between account balance computations.

PR #154, opened 2025-04-17
Define reconciliation data entity

AlexTheWizardL:
This PR goal is to define reconciliation data entity.

### DIFF
--- a/pyledger/constants.py
+++ b/pyledger/constants.py
@@ -128,6 +128,20 @@ ACCOUNT_HISTORY_SCHEMA_CSV = """
 """
 ACCOUNT_HISTORY_SCHEMA = pd.read_csv(StringIO(ACCOUNT_HISTORY_SCHEMA_CSV), skipinitialspace=True)
 
+RECONCILIATION_SCHEMA_CSV = """
+    column,             dtype,                mandatory,       id
+    period,             string[python],       True,            True
+    account,            string[python],       True,            True
+    currency,           string[python],       False,           True
+    profit_center,      string[python],       True,            True
+    balance,            Float64,              False,           False
+    report_balance,     Float64,              False,           False
+    tolerance,          Float64,              False,           False
+    document,           string[python],       False,           False
+    source,             string[python],       False,           True
+"""
+RECONCILIATION_SCHEMA = pd.read_csv(StringIO(RECONCILIATION_SCHEMA_CSV), skipinitialspace=True)
+
 DEFAULT_PRECISION = 0.01
 
 DEFAULT_ASSETS = enforce_schema(
--- a/pyledger/constants.py
+++ b/pyledger/constants.py
@@ -133,7 +133,7 @@ RECONCILIATION_SCHEMA_CSV = """
     period,             string[python],       True,            True
     account,            string[python],       True,            True
     currency,           string[python],       False,           True
-    profit_center,      string[python],       True,            True
+    profit_center,      string[python],       False,           True
     balance,            Float64,              False,           False
     report_balance,     Float64,              False,           False
     tolerance,          Float64,              False,           False
--- a/pyledger/standalone_ledger.py
+++ b/pyledger/standalone_ledger.py
@@ -6,6 +6,7 @@ independently of third-party software.
 import datetime
 import numpy as np
 import pandas as pd
+from pyledger.storage_entity import AccountingEntity
 from .decorators import timed_cache
 from .constants import JOURNAL_SCHEMA
 from .ledger_engine import LedgerEngine
@@ -19,6 +20,13 @@ class StandaloneLedger(LedgerEngine):
     accounting software. It serves as a base for any standalone ledger implementation
     with a specific data storage choice.
     """
+
+    # ----------------------------------------------------------------------
+    # Storage entities
+
+    @property
+    def reconciliation(self) -> AccountingEntity:
+        return self._reconciliation
 
     # ----------------------------------------------------------------------
     # Tax Codes
--- a/pyledger/standalone_ledger.py
+++ b/pyledger/standalone_ledger.py
@@ -42,7 +42,7 @@ class StandaloneLedger(LedgerEngine):
             self.reconciliation.mirror(reconciliation, delete=True)
 
     def clear(self):
-        """Extend to clear all data, including reconciliation records."""
+        """Extend clear() to also delete reconciliation records."""
         super().clear()
         self.reconciliation.mirror(None, delete=True)
 
--- a/pyledger/standalone_ledger.py
+++ b/pyledger/standalone_ledger.py
@@ -29,14 +29,11 @@ class StandaloneLedger(LedgerEngine):
             archive.writestr('reconciliation.csv', self.reconciliation.list().to_csv(index=False))
 
     def restore_from_zip(self, archive_path: str):
-        """Extend restore_from_zip to restore reconciliation data from archive if available."""
-        reconciliation = None
+        """Extend restore_from_zip to also restore reconciliation data."""
+        super().restore_from_zip(archive_path)
         with zipfile.ZipFile(archive_path, 'r') as archive:
             if 'reconciliation.csv' in archive.namelist():
-                reconciliation = pd.read_csv(archive.open('reconciliation.csv'))
-        super().restore_from_zip(archive_path)
-        if reconciliation is not None:
-            self.reconciliation.mirror(reconciliation, delete=True)
+                self.restore(reconciliation=pd.read_csv(archive.open('reconciliation.csv')))
 
     def restore(self, *args, reconciliation: pd.DataFrame | None = None, **kwargs):
         """Extend restore to reconciliation data after base restoration."""
--- a/pyledger/standalone_ledger.py
+++ b/pyledger/standalone_ledger.py
@@ -4,6 +4,7 @@ independently of third-party software.
 """
 
 import datetime
+import zipfile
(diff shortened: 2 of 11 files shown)


pyledger/constants.py
```diff
@@ -128,6 +128,20 @@
 """
 ACCOUNT_HISTORY_SCHEMA = pd.read_csv(StringIO(ACCOUNT_HISTORY_SCHEMA_CSV), skipinitialspace=True)
 
+RECONCILIATION_SCHEMA_CSV = """
+    column,             dtype,                mandatory,       id
+    period,             string[python],       True,            True
+    account,            string[python],       True,            True
+    currency,           string[python],       False,           True
+    profit_center,      string[python],       True,            True
```
lasuk, line 136: The profit center column is not mandatory.

pyledger/standalone_ledger.py
```diff
@@ -20,6 +22,40 @@ class StandaloneLedger(LedgerEngine):
...
+
+    def restore_from_zip(self, archive_path: str):
+        """Extend restore_from_zip to restore reconciliation data from archive if available."""
+        reconciliation = None
+        with zipfile.ZipFile(archive_path, 'r') as archive:
+            if 'reconciliation.csv' in archive.namelist():
+                reconciliation = pd.read_csv(archive.open('reconciliation.csv'))
+        super().restore_from_zip(archive_path)
+        if reconciliation is not None:
+            self.reconciliation.mirror(reconciliation, delete=True)
```
lasuk, line 39: Proposal for a streamlined version:
```
def restore_from_zip(self, archive_path: str):
    """Extend restore_from_zip to also restore reconciliation data."""
    super().restore_from_zip(archive_path)
    with zipfile.ZipFile(archive_path, 'r') as archive:
        if 'reconciliation.csv' in archive.namelist():
            self.restore(reconciliation=pd.read_csv(archive.open('reconciliation.csv')))
```

pyledger/standalone_ledger.py
```diff
@@ -20,6 +22,40 @@ class StandaloneLedger(LedgerEngine):
...
+
+    def restore(self, *args, reconciliation: pd.DataFrame | None = None, **kwargs):
+        """Extend restore to reconciliation data after base restoration."""
+        super().restore(*args, **kwargs)
+        if reconciliation is not None:
+            self.reconciliation.mirror(reconciliation, delete=True)
+
+    def clear(self):
+        """Extend to clear all data, including reconciliation records."""
```
lasuk, line 48: Alternative suggestion for docstring:
```
"""Extend clear() to also delete reconciliation records."""
```

pyledger/tests/base_test.py
```diff
@@ -138,6 +138,18 @@
 """
 PROFIT_CENTERS = pd.read_csv(StringIO(PROFIT_CENTERS_CSV), skipinitialspace=True)
 
+# TODO: Populate with real balances while working on #147
+RECONCILIATION_CSV = """
+    period,         account, currency,          profit_center,       balance,  report_balance, tolerance, document
+    2023-12-31,   1000:2999,      CHF,                       ,          0.00,            0.00,          , 2023/reconciliation/2023-12-31.pdf
+    2024-01-23,   1000:2999,      EUR,                       ,          0.00,            0.00,      0.01, 2024/reconciliation/2024-01-23.pdf
+    2024-09-25,        1000,      EUR,                       ,          0.00,            0.00,         1,
+    2024-Q4,      1000:2999,         ,                       ,          0.00,            0.00,      0.01, 2024/reconciliation/2024-Q4.pdf
```
lasuk, line 147: Line 147 is expected to be discarded (balance is defined and currency is missing).

pyledger/tests/base_test_reconciliation.py
```diff
@@ -0,0 +1,151 @@
...
+
+    @abstractmethod
+    @pytest.fixture
+    def engine(self):
+        pass
+
+    def test_reconciliation_accessor_mutators(self, engine, ignore_row_order=False):
+        # Add reconciliation one by one and with multiple rows
+        reconciliation = self.RECONCILIATION.sample(frac=1).reset_index(drop=True)
+        for revaluation in reconciliation.head(-3).to_dict('records'):
```
lasuk, line 21: Why name the variable `revaluation`?

pyledger/tests/base_test_reconciliation.py
```diff
@@ -0,0 +1,151 @@
...
+
+        # Modify all columns from the schema in a specific row
+        reconciliation.loc[3, "document"] = "Modify with all columns test"
+        engine.reconciliation.modify([reconciliation.loc[3]])
+        assert_frame_equal(
+            engine.reconciliation.list(), reconciliation,
+            check_like=True, ignore_row_order=ignore_row_order
+        )
+
+        # Modify with a multiple rows
```
lasuk, line 52: Should be 
```
# Modify multiple rows
```
Instead of
```
# Modify with a multiple rows
```

pyledger/tests/test_text_ledger_reconciliation.py
```diff
@@ -0,0 +1,66 @@
...
+
+import pandas as pd
+import pytest
+from pyledger.constants import RECONCILIATION_SCHEMA
+from .base_test_reconciliation import BaseTestReconciliation
+from pyledger import TextLedger
+from consistent_df import assert_frame_equal
+
+
+class TestRevaluations(BaseTestReconciliation):
```
lasuk, line 11: Why name the class `TestRevaluations`?

pyledger/text_ledger.py
```diff
@@ -196,7 +203,43 @@ def format_with_precision(series: pd.Series, precision: float) -> pd.Series:
...
+        df = enforce_schema(df, RECONCILIATION_SCHEMA, sort_columns=True, keep_extra_columns=True)
+
+        # Apply the smallest precision
+        def format_with_precision(series: pd.Series, precision: float) -> pd.Series:
+            """Formats a series to a specific decimal precision."""
+            decimal_places = -1 * math.floor(math.log10(precision))
+            return series.apply(lambda x: pd.NA if pd.isna(x) else f"{x:.{decimal_places}f}")
+
+        df["balance"] = format_with_precision(df["balance"], 0.01)
+        df["report_balance"] = format_with_precision(df["report_balance"], 0.01)
```
lasuk, line 231: Precision should not be hard-coded (0.01), but inferred from asset configuration.

pyledger/text_ledger.py
```diff
@@ -196,7 +203,43 @@ def format_with_precision(series: pd.Series, precision: float) -> pd.Series:
...
+            df (pd.DataFrame): The reconciliation entries to save.
+            file (str): Path of the CSV file to write.
+
+        Returns:
+            pd.DataFrame: The formatted DataFrame saved to the file.
+        """
+        df = enforce_schema(df, RECONCILIATION_SCHEMA, sort_columns=True, keep_extra_columns=True)
+
+        # Apply the smallest precision
+        def format_with_precision(series: pd.Series, precision: float) -> pd.Series:
```
lasuk, line 225: Is `format_with_precision` already defined elsewhere? If so, consider using a single common definition.

lasuk (commented):
Thanks for implementing the new reconciliation storage entity. Cool to see how convincingly our modular system can be extended.

pyledger/text_ledger.py
AlexTheWizardL: This method was previously used in `write_journal_file()`.  
For now, I've made `format_with_precision()` a static method of `TextLedger`.  
We can reconsider moving it elsewhere if it turns out to be needed in other parts of the codebase.

pyledger/text_ledger.py
```diff
@@ -196,7 +205,43 @@ def format_with_precision(series: pd.Series, precision: float) -> pd.Series:
...
+            file (str): Path of the CSV file to write.
+
+        Returns:
+            pd.DataFrame: The formatted DataFrame saved to the file.
+        """
+        df = enforce_schema(df, RECONCILIATION_SCHEMA, sort_columns=True, keep_extra_columns=True)
+        if not df.empty:
+            increment = df.apply(
+                lambda row: DEFAULT_PRECISION if pd.isna(row["currency"]) else self.precision(
+                    row["currency"], parse_date_span(row["period"])[1]), axis=1
```
lasuk, line 228: Can `parse_date_span(row["period"])` return a missing value as period start?
If you, you may consider using the period end instead.
If period end can also be missing, you need to come up with a fallback behaviour for this case, e.g. use today's date.

pyledger/text_ledger.py
lasuk: Excellent. Thanks!

pyledger/text_ledger.py
AlexTheWizardL: `parse_date_span(row["period"])[1])` condition access the period end (using `[1]` part).
Period end by definition of `parse_date_span()` method is always present, even if `row["period"]` is `NA`

AlexTheWizardL: Forgot, what I've said before. If a date is none `parse_date_span()` can return both `NA` values, so I've added a helper method to get the date in the [7ac1ebc](https://github.com/macxred/pyledger/pull/154/commits/7ac1ebcc767b449ce5c19adbbb70104f4081f63a) commit.

lasuk: Thanks for clarifying.

Funny that I still mix up Python's 0-based and R's 1-based indexing :)

PR #155, opened 2025-04-23
Add `sanitize_reconciliation()` method

AlexTheWizardL:
This PR goal is to add sanitize reconciliation method.

### DIFF
--- a/pyledger/ledger_engine.py
+++ b/pyledger/ledger_engine.py
@@ -21,6 +21,7 @@ from .constants import (
     ASSETS_SCHEMA,
     JOURNAL_SCHEMA,
     PRICE_SCHEMA,
+    RECONCILIATION_SCHEMA,
     REVALUATION_SCHEMA,
     TAX_CODE_SCHEMA,
     DEFAULT_ASSETS,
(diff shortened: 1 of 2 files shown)


lasuk:
[`sanitize_journal`](https://github.com/macxred/pyledger/blob/5e70913/pyledger/ledger_engine.py#L881-L890) implements a better readable and computationally more efficient way to handle a complex sanitation. IMO, this is a template that `sanitize_reconciliation` should follow. 

I suggest to leave the current workflow for now in the interest of wrapping up a first version of the sanitize functionality and open an issue to refactor `sanitize_reconciliation` at a later point in time.

AlexTheWizardL:
> [`sanitize_journal`](https://github.com/macxred/pyledger/blob/5e70913/pyledger/ledger_engine.py#L881-L890) implements a better readable and computationally more efficient way to handle a complex sanitation. IMO, this is a template that `sanitize_reconciliation` should follow.
> 
> I suggest to leave the current workflow for now in the interest of wrapping up a first version of the sanitize functionality and open an issue to refactor `sanitize_reconciliation` at a later point in time.

Created issue #161

PR #156, opened 2025-04-28
Add reconcile logic

AlexTheWizardL:
This PR goal is to add reconcile logic.

### DIFF
--- a/pyledger/ledger_engine.py
+++ b/pyledger/ledger_engine.py
@@ -1774,12 +1774,11 @@ class LedgerEngine(ABC):
         period: str | datetime.date | None = None,
         source_pattern: str | None = None
     ) -> pd.DataFrame:
-        """Filter and enrich reconciliation data with actual account balances.
+        """Enrich reconciliation data with actual account balances.
 
         Args:
             df (pd.DataFrame): Reconciliation input with RECONCILIATION_SCHEMA.
-            period (datetime.date | str | None):
-                The time period to filter reconciliation data.
+            period (datetime.date | str | None): Time period to filter reconciliation data.
                 See `parse_date_span` for possible values.
             source_pattern (str | None): Regex to filter rows by 'source' path.
 
--- a/pyledger/ledger_engine.py
+++ b/pyledger/ledger_engine.py
@@ -1777,10 +1777,10 @@ class LedgerEngine(ABC):
         """Enrich reconciliation data with actual account balances.
 
         Args:
-            df (pd.DataFrame): Reconciliation input with RECONCILIATION_SCHEMA.
+            df (pd.DataFrame): Expected balances compatible with RECONCILIATION_SCHEMA.
             period (datetime.date | str | None): Time period to filter reconciliation data.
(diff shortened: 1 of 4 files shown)


pyledger/ledger_engine.py
```diff
@@ -1738,3 +1739,42 @@ def assign_default_tolerance(row):
...
+    def reconcile(
+        self,
+        df: pd.DataFrame,
+        period: str | datetime.date | None = None,
+        source_pattern: str | None = None
+    ) -> pd.DataFrame:
+        """Enrich reconciliation data with actual account balances.
+
+        Args:
+            df (pd.DataFrame): Reconciliation input with RECONCILIATION_SCHEMA.
```
lasuk, line 1752: Proposal:
```
df (pd.DataFrame): Expected balances compatible with RECONCILIATION_SCHEMA.
```
Instead of:
```
df (pd.DataFrame): Reconciliation input with RECONCILIATION_SCHEMA.
```

pyledger/ledger_engine.py
```diff
@@ -1738,3 +1739,42 @@ def assign_default_tolerance(row):
...
+            period (datetime.date | str | None): Time period to filter reconciliation data.
+                See `parse_date_span` for possible values.
+            source_pattern (str | None): Regex to filter rows by 'source' path.
```
lasuk, line 1755: Proposal:
```
source_pattern (str | None): Regex for filtering the 'source' column.
```
Instead of:
```
source_pattern (str | None): Regex to filter rows by 'source' path.
```

pyledger/ledger_engine.py
```diff
@@ -1738,3 +1739,42 @@ def assign_default_tolerance(row):
...
+
+        Returns:
+            pd.DataFrame: Enriched reconciliation DataFrame with added
+                'actual_balance' and 'actual_report_balance' columns.
+        """
+        if period is not None:
+            start, end = parse_date_span(period)
+            start = pd.to_datetime(start) if start else None
```
lasuk, line 1763: Why not also convert `end` with `pd.to_datetime()`?

pyledger/ledger_engine.py
```diff
@@ -1738,3 +1739,42 @@ def assign_default_tolerance(row):
...
+
+        if source_pattern is not None:
+            df = df[df['source'].str.contains(source_pattern, na=False)]
+
+        df = self.sanitize_reconciliation(df)
+        balances = self.account_balances(df).rename(columns={
+            "balance": "actual_balance",
+            "report_balance": "actual_report_balance"
+        })
+        return pd.concat([df, balances.reset_index(drop=True)], axis=1)
```
lasuk, line 1780: I think this is more robust:

```
result = pd.concat([df.reset_index(drop=True), balances.reset_index(drop=True)], axis=1)
result.index = df.index
return result
```

Instead of:
```
 return pd.concat([df, balances.reset_index(drop=True)], axis=1)
```

The proposed version also works, if indexes of pd and of concat are only partially overlapping. 
(No need to state that I find indexes overly complicated).

PR #170, opened 2025-05-19
Add `parse_profit_center_range()` method

AlexTheWizardL:


AlexTheWizardL:
I've implemented all suggested changes except one.
We don't have the `sanitize_profit_centers()` method yet. I propose to create an issue to define one.

lasuk:
> We don't have the sanitize_profit_centers() method yet. I propose to create an issue to define one.

Agreed. Thanks.

pyledger/tests/test_parse_profit_center_range.py
```diff
@@ -0,0 +1,60 @@
...
+)
+def test_parse_profit_centers_valid_inputs(input_value, expected_output, engine):
+    assert engine.parse_profit_centers(input_value) == expected_output
+
+
+@pytest.mark.parametrize(
+    "invalid_input",
+    [
+        {"add": ["Shop", 123]},  # dicts no longer allowed
+        {"subtract": ["General"]},
```
lasuk, line 39: Testing for dicts seems unimportant. I would drop these two test cases.

pyledger/ledger_engine.py
```diff
@@ -1889,3 +1889,52 @@ def account_description(account):
...
+                - list[str]: A list of profit center names.
+
+        Returns:
+            set[str]: A set of valid profit center names.
+
+        Raises:
+            ValueError: If the input format is invalid or no valid profit centers are found.
+        """
+        if isinstance(profit_center, list):
+            items = list(profit_center)
```
lasuk, line 1915: Why convert to a list? I suggest
```
items = set(profit_center)
```

Instead of:
```
items = list(profit_center)
```

pyledger/ledger_engine.py
```diff
@@ -1889,3 +1889,52 @@ def account_description(account):
...
+                - str: A string with profit centers separated by '+'.
+                - list[str]: A list of profit center names.
+
+        Returns:
+            set[str]: A set of valid profit center names.
+
+        Raises:
+            ValueError: If the input format is invalid or no valid profit centers are found.
+        """
+        if isinstance(profit_center, list):
```
lasuk, line 1914: I would also accept sets as input.

PR #177, opened 2025-06-09
Refactor account balance methods

AlexTheWizardL:


pyledger/standalone_ledger.py
```diff
@@ -273,29 +283,34 @@ def _balance_from_serialized_ledger(
...
+            return {"reporting_currency": 0.0}
+
+        sub = ledger.loc[rows, ["account", "amount", "report_amount", "currency"]]
+        sub = sub.merge(multipliers, on="account", how="inner")
+        sub["amount"] *= sub["multiplier"]
+        sub["report_amount"] *= sub["multiplier"]
+
+        account_currencies = sub["account"].map(lambda acc: self.account_currency(acc))
+        sub.loc[account_currencies == self.reporting_currency, "amount"] = sub["report_amount"]
+        sub["currency"] = account_currencies.combine_first(sub["currency"])
```
lasuk, line 302: Why do we need the account currency? Our preferred way to infer currencies is from (ledger) transactions, not via the account map.

lasuk (commented):
Thanks for refactoring account_balance. I left one comment.

PR #180, opened 2025-06-13
Add `sanitize_profit_centers()` method and test

AlexTheWizardL:


pyledger/ledger_engine.py
```diff
@@ -1834,6 +1835,28 @@ def account_description(account):
     # ----------------------------------------------------------------------
     # Profit center
 
+    def sanitize_profit_center(self, df: pd.DataFrame) -> pd.DataFrame:
+        """Discard profit centers that contain the '+' character and log a warning.
+        The '+' character is reserved for profit center concatenation.
```
lasuk, line 1840: Please add a blank line to separate doc string title from the body:
```
"""Discard profit centers that contain the '+' character and log a warning.

The '+' character is reserved for profit center concatenation.
```

Instead of:
```
"""Discard profit centers that contain the '+' character and log a warning.
The '+' character is reserved for profit center concatenation.
```

PR #186, opened 2025-06-27
Vectorize precision lookup logic

AlexTheWizardL:


AlexTheWizardL:
**Initial profiling result for `precision()`**
Total time: **0.183142 s**
Function: `precision` (line 1412)

**After refactor (now entry point is `precision_vectorized`, cache removed):**
Total time: **1.73235 s**
Function: `precision_vectorized` (line 1422)

Despite the refactor, performance hasn’t improved—likely because the method is called 2,030 times, according to the profiler.
This is addressed in the next PR #187.

lasuk:
Thanks a lot for converting the precision functionality to a unified vectorised interface. 
This is very helpful for further performance improvements of our code base.!

PR #195, opened 2025-07-24
Add `report_balance()` method

AlexTheWizardL:


lasuk:
@AlexTheWizardL: Linter is not happy. Can you have a look? Thanks.

pyledger/ledger_engine.py
```diff
@@ -15,6 +15,8 @@
 import openpyxl
 import pandas as pd
 import polars as pl
+from .reporting import summarize_groups
+from .typst import format_threshold, df_to_typst
```
lasuk, line 19: I propose to remove `format_threshold` from module typst.py. See #198.

pyledger/ledger_engine.py
```diff
@@ -2023,3 +2025,123 @@ def _invalid_lookup_profit_centers(self, df: pd.DataFrame, invalid_ids: set) ->
...
+            summarize={col: "sum" for col in label_cols},
+            group="group",
+            description="description",
+            leading_space=1,
+            staggered=staggered
+        )
+
+        threshold = self.precision_vectorized([self.reporting_currency], [None])[0] / 2
+        for col in label_cols:
+            report[col] = format_threshold(report[col], threshold=threshold)
```
lasuk, line 2130: Alternative:
```
report[col] = [report[col] if abs(report[col]) < threshold else pd.NA]
```
Instead of:
```
report[col] = format_threshold(report[col], threshold=threshold)
```

pyledger/ledger_engine.py
```diff
@@ -2023,3 +2025,123 @@ def _invalid_lookup_profit_centers(self, df: pd.DataFrame, invalid_ids: set) ->
...
+                - "typst": Formatted string for Typst rendering.
+                - "dataframe": Raw DataFrame for inspection or export.
+
+        Returns:
+            str or pd.DataFrame:
+                - Typst string if `format="typst"` (for PDF layout).
+                - DataFrame if `format="dataframe"` (for further processing).
+        """
+        dfs = [self._report_column(row, accounts, prune_level) for row in config.to_dict("records")]
+        sheet = pd.concat(dfs, axis=1).loc[:, ~pd.concat(dfs, axis=1).columns.duplicated()]
```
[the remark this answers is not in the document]
AlexTheWizardL: Used `.loc[:, ~...duplicated()]` here because each `_report_column` call returns `"group"` and `"description"` - we only need one copy across all columns. Not about suppressing label collisions, just avoiding redundant metadata.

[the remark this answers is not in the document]
lasuk: Thanks. Understood. I would never have guessed the purpose from the code.

Please make this line easier to understand, either by explicitly naming columns to drop, or by adding your explanation as a comment.

pyledger/ledger_engine.py
lasuk: A centralised `format_number` hard-codes number formatting. I suggest to define a `format_number` argument instead, with default: 
```
lambda x: f"{x:,.2f}"
```

lasuk:
This PR was accidentally closed by merging #191. Re-opening now.

pyledger/ledger_engine.py
AlexTheWizardL: Used `def` instead of `lambda` to avoid flake8 E731.

PR #205, opened 2025-08-07
Add support for ledger entries source tracking via `source_column` in `TextLedger`.

AlexTheWizardL:


lasuk:
Dropped commit 2cf5d93: feat: change id for revaluation and target balance entries

I am not sure we should change the ID of auto-generated journal entries. Let's discuss this in our next call.

lasuk:
Decided to leave commit 2cf5d (feat: change id for revaluation and target balance entries) aside for now.
- It is not part of the current version of the PR.
- We can add the functionality later if it turns out to be helpful.

PR #237, opened 2025-10-06
Fix Typst special character escaping

AlexTheWizardL:
Fixes #226


AlexTheWizardL:
Extended `escape_typst_text()` to escape all Typst special characters: \ $ # * @ < >

Added `raw` parameter to `report_table()` and `account_sheet_tables()` to optionally disable escaping for users who need Typst markup in their data.

Added comprehensive tests for both functions.

